PR #11, opened 2023-02-28
decrease the 'phase' values

GenialHacker:
In [this commit](https://github.com/coreruleset/coreruleset/commit/5a47465f9cee33d46ac62b03fd300195bbee8c2f) made three years ago, the 'phase' values for some rules were reduced, including the rule [911100](https://github.com/coreruleset/coreruleset/commit/5a47465f9cee33d46ac62b03fd300195bbee8c2f#diff-74c2e602499bf7b7a23dbbe1bbb2ba1251c466e60ab3c3684210bb2aef042773) which is responsible for handling request methods. The NextCloud DAV functionality doesn't work with the previous values.

PR #3, opened 2022-04-30
fix: Adjust phases for rules setting TX variables

theseion:
TX variables in phase 1 are overwritten by the CRS initialization. Moving them to phase 2 does *not* fix the issue but is in line with the how the rules were executed before they were moved into the plugin.

Relates to #2

PR #12, opened 2023-03-06
fix FPs

GenialHacker:
FPs at PL 1.

### DIFF
--- a/plugins/nextcloud-rule-exclusions-before.conf
+++ b/plugins/nextcloud-rule-exclusions-before.conf
@@ -160,6 +160,17 @@ SecRule REQUEST_METHOD "@streq REPORT" \
         "t:none,\
         ctl:ruleRemoveById=920340"
 
+# FP when NextCloud default app "Text" detects text files in file manager.
+# PUT - When the "Text" app tries to create a session in file manager. 
+SecRule REQUEST_FILENAME "@rx /apps/text/session/create" \
+    "id:9508122,\
+    phase:1,\
+    pass,\
+    t:none,\
+    nolog,\
+    ver:'nextcloud-rule-exclusions-plugin/1.0.0',\
+    setvar:'tx.allowed_methods=%{tx.allowed_methods} PUT'"
+
 
 #
 # [ Searchengine ]
@@ -477,7 +488,20 @@ SecRule REQUEST_FILENAME "@contains /.well-known/carddav" \
     ver:'nextcloud-rule-exclusions-plugin/1.0.0',\
     setvar:'tx.allowed_methods=%{tx.allowed_methods} PROPFIND'"
 
+# Appearance and accessibility
 
+# Enabling any default theme requires the PUT method.
+# This FP prevents any theme from permanent activation.
+SecRule REQUEST_FILENAME "@rx /ocs/v[0-9]+\.php/apps/theming/api/v[0-9]+/theme/.*?/enable" \
+    "id:9508604,\
+    phase:1,\
+    pass,\
+    t:none,\
+    nolog,\
+    ver:'nextcloud-rule-exclusions-plugin/1.0.0',\
+    setvar:'tx.allowed_methods=%{tx.allowed_methods} PUT'"
+    
+    
 #
 # [ Notifications ]
 #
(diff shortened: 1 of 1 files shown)


azurit:
@GenialHacker Hi and thanks! Can you look at my comment in the code?

GenialHacker:
Hi @azurit, you're right! I don't remember why I did that, but I changed it as you said.

plugins/nextcloud-rule-exclusions-before.conf
```diff
@@ -477,7 +488,20 @@ SecRule REQUEST_FILENAME "@contains /.well-known/carddav" \
     ver:'nextcloud-rule-exclusions-plugin/1.0.0',\
     setvar:'tx.allowed_methods=%{tx.allowed_methods} PROPFIND'"
 
+# Appearance and accessibility
 
+# Enabling any default theme requires the PUT method.
+# This FP prevents any theme from permanent activation.
```
azurit, line 494: Please edit this comment. Don't write about FPs but the opposite, write what rule does. Anyway, I think the first sentence is enough.

GenialHacker: Updated.

azurit (approved):
Looks good!

azurit:
@GenialHacker Sorry i noticed one more thing: We need to move all rules into phase 2 because of the TX variables (for plugins, they are not initialized in phase 1).

GenialHacker:
Yeah, I also thought about that when writing those rules. But, if you look at my last PR #11, then you'll notice that rule causing this FP, i.e rule 911100 was moved to phase 1 three years ago. So, the FPs won't stop if you move the rules in phase 2.

GenialHacker:
If you want me create a PR moving the rule 911100 to phase 2, then I'll be really glad to do that. :)

azurit:
But all of them should be back to phase 2, see this:
https://github.com/coreruleset/nextcloud-rule-exclusions-plugin/pull/3/files

@theseion Any suggestions how to resolve this?

GenialHacker:
Looks like my last PR only fixed two of those and the rest also needs to be put back into phase 1.

azurit:
It won't help as TX variables are initialized in phase 1 AFTER the plugins are executed, so changes done by plugins gets overwritten.

azurit:
One possible solution for this problem would be to not modify `TX` variables but exclude the whole rule using `ctl:ruleRemoveById`.

GenialHacker:
I can't see how that's going to help. I always thought `setvar:'tx.allowed_methods` is the right and elegant solution. Anyways, it's @theseion's call.

theseion:
Oh boy. As I discovered in your other PR @azurit, at least some of the rules are actually required to run in phase 1.

I think we need to fix this properly, once and for all. The way to do that, I think, is to move most rules into a new `*-after.conf` file. That will ensure that variables have been initialised. Does that sound reasonable @azurit?

azurit:
@theseion Will it really help for example for rule 911100 which runs in phase 1?

theseion:
No, I just realised that too. If we want to be able to do this like we want to, then we will need a mechanism to run the rule _after_ initialisation but _before_ method enforcement rules.

Let me open up a separate issue to discuss this, as it affects the entire plugin system.

theseion:
As the rules are running in phase 1 at the moment and we don't have an immediate fix for the issue, I suggest that you still merge this PR, targeting phase 1. If we then start moving things around those rules will be included in the move.

GenialHacker:
I think we should first change other rules that were moved to phase 2 in PR #3. Or those are okay with you guys?

azurit:
> If we want to be able to do this like we want to, then we will need a mechanism to run the rule _after_ initialisation but _before_ method enforcement rules.

We were already discussing this but @dune73 was not happy with this solution because it will complicate the CRS includes/installation. But i agree that this is the best option.

azurit:
> I think we should first change other rules that were moved to phase 2 in PR #3. Or those are okay with you guys?

Yes, for me.

GenialHacker:
@azurit Can you look into [this PR](https://github.com/coreruleset/coreruleset/pull/3155)?

azurit:
Looks good, merging!

PR #14, opened 2023-03-21
feat: Copy of tx.allowed_request_content_type initialization

azurit:
As discussed on Slack, this is needed for proper modification of `tx.allowed_request_content_type` by exclusion rules. Plugins are currently running before CRS initialization so variables like `tx.allowed_request_content_type` needs to be initialized also by plugins if they are going to be modified.

### DIFF
--- a/plugins/nextcloud-rule-exclusions-before.conf
+++ b/plugins/nextcloud-rule-exclusions-before.conf
@@ -50,7 +50,7 @@ SecRule TX:nextcloud-rule-exclusions-plugin_enabled "@eq 0" "id:9508099,phase:1,
 
 # Copy of default HTTP policy: allowed_request_content_type (rule 900220 in crs-setup.conf)
 # We need to initialize this also here because plugin setup runs before CRS initialization (this
-# is a know limitation of current plugin implementation). Must be kept in sync with CRS
+# is a known limitation of the current plugin architecture). Must be kept in sync with CRS
 # default setting.
 SecRule &TX:allowed_request_content_type "@eq 0" \
     "id:9508100,\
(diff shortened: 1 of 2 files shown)


azurit:
@dune73 Can you be so kind and look at this? Is this done by the way we agreed yesterday on Slack?

plugins/nextcloud-rule-exclusions-before.conf
```diff
@@ -48,13 +48,26 @@ SecRule TX:nextcloud-rule-exclusions-plugin_enabled "@eq 0" "id:9508099,phase:1,
 # in libmodsecurity3!
 
 
+# Copy of default HTTP policy: allowed_request_content_type (rule 900220 in crs-setup.conf)
+# We need to initialize this also here because plugins runs before CRS initialization (this
```
theseion, line 52:
```suggestion
# We need to initialize this also here because plugin setup runs before CRS initialization (this
```

plugins/nextcloud-rule-exclusions-before.conf
```diff
@@ -48,13 +48,26 @@ SecRule TX:nextcloud-rule-exclusions-plugin_enabled "@eq 0" "id:9508099,phase:1,
...
+# is a know limitation of current plugin implementation). Must be kept in sync with CRS
```
theseion, line 53:
```suggestion
# is a known limitation of the current plugin architecture). Must be kept in sync with CRS
```

theseion (changes requested):

theseion:
We need the same for HTTP methods. Are you going to open that one too?

azurit:
Yes i just need an confirmation that this is the right way to do it.

theseion:
I believe it is. That was my take away from the discussion.

theseion (approved):

dune73:
Sorry for being late to the show. I agree with the resolution.

PR #15, opened 2023-03-28
Fix for few FPs

Xhoenix:
As promised in my last PR, here are 10 more FPs. There are a few more FPs, but I forgot to label them as PL 1 and they got mixed with my PL 2 FPs. @theseion if you find a typo or mistake, kindly make changes where necessary. Thank you.

Also, we need to update the plugin version to **1.0.1**, as done with WordPress plugin.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -16,3 +16,4 @@
 - [Jozef Sudolský](https://github.com/azurit)
 - [Timo](https://github.com/ntimo)
 - [Felipe Zipitría](https://github.com/fzipi)
+- [Jitendra Patro](https://github.com/Xhoenix)
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -15,5 +15,5 @@
 - [Federico G. Schwindt](https://github.com/fgsch)
 - [Jozef Sudolský](https://github.com/azurit)
 - [Timo](https://github.com/ntimo)
+- [Xhoenix](https://github.com/Xhoenix)
 - [Felipe Zipitría](https://github.com/fzipi)
-- [Jitendra Patro](https://github.com/Xhoenix)
--- a/plugins/nextcloud-rule-exclusions-before.conf
+++ b/plugins/nextcloud-rule-exclusions-before.conf
@@ -685,7 +685,7 @@ SecRule REQUEST_FILENAME "@rx /ocs/v[0-9]+\.php/profile" \
 # [ Survey Client ]
 #
 
-# FP when trying to reject the survey client
+# When trying to reject metrics sharing
 SecRule REQUEST_FILENAME "@rx /ocs/v[0-9]+\.php/apps/survey_client/api" \
     "id:9508860,\
     phase:1,\
--- a/plugins/nextcloud-rule-exclusions-before.conf
+++ b/plugins/nextcloud-rule-exclusions-before.conf
@@ -700,7 +700,7 @@ SecRule REQUEST_FILENAME "@rx /ocs/v[0-9]+\.php/apps/survey_client/api" \
 #
 
 # When trying to remove account from Mobile App
-SecRule REQUEST_FILENAME "@rx /ocs/v[0-9]+.php/core/apppassword" \
+SecRule REQUEST_FILENAME "@rx /ocs/v[0-9]+\.php/core/apppassword" \
     "id:9508870,\
     phase:1,\
     pass,\
--- a/plugins/nextcloud-rule-exclusions-before.conf
+++ b/plugins/nextcloud-rule-exclusions-before.conf
@@ -700,7 +700,7 @@ SecRule REQUEST_FILENAME "@rx /ocs/v[0-9]+\.php/apps/survey_client/api" \
 #
 
 # When trying to remove account from Mobile App
-SecRule REQUEST_FILENAME "@contains /ocs/v2.php/core/apppassword" \
+SecRule REQUEST_FILENAME "@rx /ocs/v[0-9]+.php/core/apppassword" \
     "id:9508870,\
     phase:1,\
     pass,\
--- a/plugins/nextcloud-rule-exclusions-before.conf
+++ b/plugins/nextcloud-rule-exclusions-before.conf
@@ -507,7 +507,7 @@ SecRule REQUEST_FILENAME "@contains /settings/personal/authtokens/" \
 
 # When trying to update availability in Personal --> Availability
 SecRule REQUEST_FILENAME "@contains /remote.php/dav/calendars/admin/inbox" \
-    "id:9508602,\
+    "id:9508606,\
     phase:1,\
     pass,\
     t:none,\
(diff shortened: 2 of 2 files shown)


plugins/nextcloud-rule-exclusions-before.conf
```diff
@@ -526,6 +558,17 @@ SecRule REQUEST_FILENAME "@contains /.well-known/carddav" \
     ver:'nextcloud-rule-exclusions-plugin/1.0.0',\
     setvar:'tx.allowed_methods=%{tx.allowed_methods} PROPFIND'"
 
+# Administration --> Sharing -- Federated cloud sharing
```
theseion, line 561:
```suggestion
# Administration --> Sharing --> Federated cloud sharing
```

plugins/nextcloud-rule-exclusions-before.conf
```diff
@@ -505,6 +505,38 @@ SecRule REQUEST_FILENAME "@contains /settings/personal/authtokens/" \
...
+SecRule REQUEST_FILENAME "@contains /apps/firstrunwizard/wizard" \
+    "id:9508603,\
+    phase:1,\
+    pass,\
+    t:none,\
+    nolog,\
+    ver:'nextcloud-rule-exclusions-plugin/1.0.0',\
+    setvar:'tx.allowed_methods=%{tx.allowed_methods} DELETE'"
+
+# When trying to update Personal --> Sharing --> Sharing --> Accept user and groups shares by default
```
theseion, line 529: This the double `--> Sharing` correct?

plugins/nextcloud-rule-exclusions-before.conf
```diff
@@ -584,19 +636,77 @@ SecRule REQUEST_FILENAME "@rx /ocs/v[0-9]+\.php/apps/user_status/api/v[0-9]+/hea
...
+    t:none,\
+    nolog,\
+    ver:'nextcloud-rule-exclusions-plugin/1.0.0',\
+    setvar:'tx.allowed_methods=%{tx.allowed_methods} PUT'"
+
+
+# [ Survey Client ]
+#
+
+# FP when trying to reject the survey client
```
theseion, line 688: Not sure whether `cancel` is better than `reject` here, as I don't know the context.
```suggestion
# When trying to cancel the survey client
```

plugins/nextcloud-rule-exclusions-before.conf
```diff
@@ -584,19 +636,77 @@ SecRule REQUEST_FILENAME "@rx /ocs/v[0-9]+\.php/apps/user_status/api/v[0-9]+/hea
...
+    t:none,\
+    nolog,\
+    ver:'nextcloud-rule-exclusions-plugin/1.0.0',\
+    setvar:'tx.allowed_methods=%{tx.allowed_methods} DELETE'"
+
+#
+# [ Mobile App ]
+#
+
+# FP when trying to remove account from Mobile App
```
theseion, line 702:
```suggestion
# When trying to remove account from Mobile App
```

plugins/nextcloud-rule-exclusions-before.conf
```diff
@@ -584,19 +636,77 @@ SecRule REQUEST_FILENAME "@rx /ocs/v[0-9]+\.php/apps/user_status/api/v[0-9]+/hea
...
+SecRule REQUEST_FILENAME "@contains /ocs/v2.php/core/apppassword" \
```
theseion, line 703: Did you also want to use a regular expression here like in the other rules?

plugins/nextcloud-rule-exclusions-before.conf
```diff
@@ -505,6 +505,38 @@ SecRule REQUEST_FILENAME "@contains /settings/personal/authtokens/" \
...
+    phase:1,\
+    pass,\
+    t:none,\
+    nolog,\
+    ver:'nextcloud-rule-exclusions-plugin/1.0.0',\
+    setvar:'tx.allowed_request_content_type=%{tx.allowed_request_content_type} |text/plain|'"
+
+
+# While going through the first run setup wizard
+SecRule REQUEST_FILENAME "@contains /apps/firstrunwizard/wizard" \
```
theseion, line 520: You could probably use `@beginsWith` instead of `@contains` for all the rules here that use `@contains`.

theseion (changes requested):

theseion:
Thanks you very much! Yes, we will increase the version number.

plugins/nextcloud-rule-exclusions-before.conf
Xhoenix: Yes, it's double `--> sharing`.

plugins/nextcloud-rule-exclusions-before.conf
Xhoenix: The Survey client pops up in a notification asking whether you wan to accept or reject the survey which'll sends statistics about your Nextcloud Installation.

theseion: I've updated the suggestion

plugins/nextcloud-rule-exclusions-before.conf
Xhoenix: The problem with `@beginsWith` is that some installations might install Nextcloud in a subfolder in relation to the webroot, and the above will look something like `/nextcloud/apps/firstrunwizard/wizard`, in which case this fix will not work. So, it's better to use `@contains` or `@endsWith` in this situation.

Xhoenix:
> Thanks you very much! Yes, we will increase the version number.

I can increase it from my side. It's a simple `sed` command.

plugins/nextcloud-rule-exclusions-before.conf
Xhoenix: I also thought about using `@rx` but then decided not to. I'll leave this as is.

Xhoenix: Okay, I realized what's wrong and changed it. Thanks.

theseion:
> I can increase it from my side. It's a simple `sed` command.

No, I'd rather have a separate PR for that.

theseion (approved):

theseion:
There's something wrong with the test setup. I need to investigate that before we can merge.

That reminds me: if you're up for it it would be nice to have your FP's covered by tests.

azurit:
Guys, i would not change the version numbers yet. None of the plugins was officially released yet and there is no need for a version change. Let's wait until CRS 4.0 and then we will release all plugins under current versions (and also we will do a downloadable release here on github).

azurit:
@Xhoenix Can you add yourself into the `CONTRIBUTORS.md`? Thanks.

CONTRIBUTORS.md
```diff
@@ -16,3 +16,4 @@
 - [Jozef Sudolský](https://github.com/azurit)
 - [Timo](https://github.com/ntimo)
 - [Felipe Zipitría](https://github.com/fzipi)
+- [Jitendra Patro](https://github.com/Xhoenix)
```
azurit, line 19: People are alphabetically ordered by surname so i suggest to move your name  before `pyllyukko`.

theseion:
@Xhoenix rule ID `9508602` is a duplicate, that's why the tests fail.

theseion:
Thanks @Xhoenix. @azurit, I'll leave merging to you.

azurit:
Merged, thanks to everyone!

Issue #16, opened 2023-04-04
Lots of FPs with NextCloud `autosaveContent`


Xhoenix:
This is the exclusion rule I'm currently using for my NextCloud Installation:-
```
SecRule REQUEST_URI "@beginsWith /apps/text/session/sync" \
    "id:1219,\
    phase:1,\
    pass,\
    t:none,\
    nolog,\
    ver:'nextcloud-rule-exclusions-plugin/1.0.0',\
    ctl:ruleRemoveTargetById=931130;ARGS:autosaveContent,\
    ctl:ruleRemoveTargetById=932200;ARGS:autosaveContent,\
    ctl:ruleRemoveTargetById=942150;ARGS:autosaveContent,\
    ctl:ruleRemoveTargetById=942410;ARGS:autosaveContent,\
    ctl:ruleRemoveTargetById=942430;ARGS:autosaveContent,\
    ctl:ruleRemoveTargetById=942440;ARGS:autosaveContent,\
    ctl:ruleRemoveTargetById=942431;ARGS:autosaveContent,\
    ctl:ruleRemoveTargetById=942460;ARGS:autosaveContent,\
    ctl:ruleRemoveTargetById=941340;ARGS:autosaveContent,\
    ctl:ruleRemoveTargetById=942550;ARGS:autosaveContent,\
    ctl:ruleRemoveTargetById=942131;ARGS:autosaveContent,\
    ctl:ruleRemoveTargetById=942200;ARGS:autosaveContent,\
    ctl:ruleRemoveTargetById=942210;ARGS:autosaveContent,\
    ctl:ruleRemoveTargetById=942300;ARGS:autosaveContent,\
    ctl:ruleRemoveTargetById=942370;ARGS:autosaveContent,\
    ctl:ruleRemoveTargetById=942380;ARGS:autosaveContent,\
    ctl:ruleRemoveTargetById=942520;ARGS:autosaveContent,\
    ctl:ruleRemoveTargetById=942490;ARGS:autosaveContent,\
    ctl:ruleRemoveTargetById=932235;ARGS:autosaveContent,\
    ctl:ruleRemoveTargetById=933120;ARGS:autosaveContent,\
    ctl:ruleRemoveTargetById=933161;ARGS:autosaveContent,\
    ctl:ruleRemoveTargetById=941330;ARGS:autosaveContent,\
    ctl:ruleRemoveTargetById=942120;ARGS:autosaveContent,\
[...]

PR #17, opened 2023-04-04
Fix few FPs

Xhoenix:
Let me know if you guys need the necessary logs for these FPs.

Also, @theseion can you look into [this issue](https://github.com/coreruleset/nextcloud-rule-exclusions-plugin/issues/16)?

### DIFF
--- a/plugins/nextcloud-rule-exclusions-before.conf
+++ b/plugins/nextcloud-rule-exclusions-before.conf
@@ -208,6 +208,17 @@ SecRule REQUEST_FILENAME "@contains /remote.php/dav/comments" \
     nolog,\
     ver:'nextcloud-rule-exclusions-plugin/1.0.0',\
     setvar:'tx.allowed_request_content_type=%{tx.allowed_request_content_type} |text/plain|'"
+    
+# When trying to change the "Show recommendations" option in file settings
+SecRule REQUEST_FILENAME "@contains /apps/recommendations/settings/enabled" \
+    "id:9508124,\
+    phase:1,\
+    pass,\
+    t:none,\
+    nolog,\
+    ver:'nextcloud-rule-exclusions-plugin/1.0.0',\
+    setvar:'tx.allowed_methods=%{tx.allowed_methods} PUT'"
+
         
         
 #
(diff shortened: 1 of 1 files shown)


theseion (approved):
LGTM

plugins/nextcloud-rule-exclusions-before.conf
```diff
@@ -536,6 +547,26 @@ SecRule REQUEST_FILENAME "@contains /apps/files_sharing/settings/defaultAccept"
     ver:'nextcloud-rule-exclusions-plugin/1.0.0',\
     setvar:'tx.allowed_methods=%{tx.allowed_methods} PUT'"
 
+# When trying to delete profile picture in Personal --> Personal info
+SecRule REQUEST_FILENAME "@contains /avatar" \
```
azurit, line 551: This one looks too wide. Is it possible to make it less generic?

Xhoenix: This is the relevant log for that part:- 


`[Tue Apr 04 22:19:42.908426 2023] [:error] [pid 56994] [client 127.0.0.1:47434] [client 127.0.0.1]`
`ModSecurity: Warning. Match of "within %{tx.allowed_methods}" against "REQUEST_METHOD" required.`
`[file "/etc/modsecurity/coreruleset/rules/REQUEST-911-METHOD-ENFORCEMENT.conf"] [line "44"]`
`[id "911100"] [msg "Method is not allowed by policy"] [data "DELETE"] [severity "CRITICAL"] [ver`
`"OWASP_CRS/4.0.0-rc1"] [tag "application-multi"] [tag "language-multi"] [tag "platform-multi"] [tag`
`"attack-generic"] [tag "paranoia-level/1"] [tag "OWASP_CRS"] [tag "capec/1000/210/272/220/274"] [tag`
`"PCI/12.1"] [hostname "localhost"] [uri "/avatar"] [unique_id "ZCxVJmQsstiQe1UUjjQ0VwAAAAI"]`


What do you think about `"@rx ^/avatar$"`?

Xhoenix: Or maybe just `"@rx /avatar$"`, in case NextCloud is installed in a sub-folder.

azurit: > Or maybe just `"@rx /avatar$"`, in case NextCloud is installed in a sub-folder.

Yes, this will be the best option BUT use `@endsWith` as it's much less expensive than regexp (the result will be the same).

azurit (approved):

azurit:
Merged, thank you!

PR #22, opened 2023-04-20
fix: update phase values for allowed content types

Xhoenix:
- Updated phase values for allowed content types
- Added one new FP


azurit (approved):

azurit:
Thank you, merging!

JitPatro:
I suggest adding a fix for the "autosave" issue in NextCloud, otherwise there'll be still a lot of FPs.

Xhoenix:
I suggest adding a fix for the "autosave" issue in NextCloud, otherwise there'll be still a lot of FPs.

PR #24, opened 2023-05-06
fix: autosave and logging FPs

Xhoenix:
The **plugin-lint** fails, but testing this on a live Installation works fine.


theseion (approved):
LGTM

theseion:
@Xhoenix please consider enabling automatic commit signing. We have a merge policy in place for that. See the docs here: https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits.

PR #26, opened 2023-05-16
Fix FP when logging in, Nextcloud Text app and when adding custom CSS code

EsadCetiner:
This PR fixes a few false positives
1. Rules for fixing false positives when logging in, changing passwords, logging out etc assume the Nextcloud instance is not using [pretty URLs](https://docs.nextcloud.com/server/latest/admin_manual/installation/source_installation.html#pretty-urls). I've removed  ``/index.php`` which should work for Nextcloud instances that both are and aren't using [pretty URLs](https://docs.nextcloud.com/server/latest/admin_manual/installation/source_installation.html#pretty-urls).
2. Fix few additional false positives with Nextcloud Text app including when accessing Nextcloud Text app via a public share.
3. Fix false positives with adding custom CSS code via Nextcloud Custom CSS app.

I've done ``ctl:ruleRemoveTargetByTag=OWASP_CRS;ARGS"`` for 9508128 and 9508129 since the parameter keeps on changing, it can vary from ``ARGS:json.steps.array_0.slice.content.array_0.content.array_0.text`` to ``ARGS:json.steps.array_100.slice.content.array_0.content.array_0.text``.

### DIFF
--- a/plugins/nextcloud-rule-exclusions-before.conf
+++ b/plugins/nextcloud-rule-exclusions-before.conf
@@ -230,7 +230,7 @@ SecRule REQUEST_URI "@contains /apps/text/session/sync" \
     ctl:ruleRemoveTargetByTag=OWASP_CRS;ARGS:autosaveContent"
 
 # Text app content can be anything
-SecRule REQUEST_FILENAME "@endsWith /apps/text/public/session/sync"
+SecRule REQUEST_FILENAME "@endsWith /apps/text/public/session/sync" \
    "id:9508127,\
     phase:1,\
     pass,\
@@ -240,7 +240,7 @@ SecRule REQUEST_FILENAME "@endsWith /apps/text/public/session/sync"
     ctl:ruleRemoveTargetByTag=OWASP_CRS;ARGS:json.autosaveContent"
 
 # Text app content can be anything
-SecRule REQUEST_FILENAME "@endsWith /apps/text/public/session/push"
+SecRule REQUEST_FILENAME "@endsWith /apps/text/public/session/push" \
    "id:9508128,\
     phase:1,\
     t:none,\
@@ -250,7 +250,7 @@ SecRule REQUEST_FILENAME "@endsWith /apps/text/public/session/push"
     ctl:ruleRemoveTargetByTag=OWASP_CRS;ARGS"
 
 # Text app content can be anything
-SecRule REQUEST_FILENAME "@endsWith /apps/text/session/push"
+SecRule REQUEST_FILENAME "@endsWith /apps/text/session/push" \
    "id:9508129,\
     phase:1,\
     pass,\
@@ -868,7 +868,7 @@ SecRule REQUEST_FILENAME "@contains /apps/notes/notes/" \
 #
 
 # Allow admin to add custom CSS code
-SecRule REQUEST_FILENAME "@rx /ocs/v[0-9]\.php/apps/provisioning_api/api/v[0-9]/config/apps/theming_customcss/customcss$"
+SecRule REQUEST_FILENAME "@rx /ocs/v[0-9]\.php/apps/provisioning_api/api/v[0-9]/config/apps/theming_customcss/customcss$" \
    "id:9508930,\
     phase:1,\
     pass,\
--- a/plugins/nextcloud-rule-exclusions-before.conf
+++ b/plugins/nextcloud-rule-exclusions-before.conf
@@ -863,4 +863,18 @@ SecRule REQUEST_FILENAME "@contains /apps/notes/notes/" \
     ctl:ruleRemoveTargetByTag=OWASP_CRS;ARGS:json.title,\
     setvar:'tx.allowed_methods=%{tx.allowed_methods} PUT DELETE'"
 
+#
+# [ Custom CSS ]
+#
+
+# Allow admin to add CSS custom CSS code
+SecRule REQUEST_FILENAME "@rx /ocs/v[0-9]\.php/apps/provisioning_api/api/v[0-9]/config/apps/theming_customcss/customcss$"
+   "id:9508930,\
+    phase:1,\
+    pass,\
+    t:none,\
+    nolog,\
+    ver:'nextcloud-rule-exclusions-plugin/1.0.0',\
+    ctl:ruleRemoveTargetByTag=OWASP_CRS;ARGS:value"    
+   
 SecMarker "END-NEXTCLOUD-ADMIN"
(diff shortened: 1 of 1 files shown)


azurit:
Hi @EsadCetiner! Thank you for this PR. Can you explain why it is needed to completely disable CRS in all of new rules?

plugins/nextcloud-rule-exclusions-before.conf
```diff
@@ -457,7 +487,7 @@ SecRule REQUEST_FILENAME "@contains /index.php/apps/bookmarks/" \
 # This removes checks on the 'password' and related fields:
 
 # User login password.
-SecRule REQUEST_FILENAME "@contains /index.php/login" \
+SecRule REQUEST_FILENAME "@contains /login" \
```
azurit, line 490: Is `@contains` really needed here? What about `@endsWith`?

plugins/nextcloud-rule-exclusions-before.conf
```diff
@@ -228,7 +228,37 @@ SecRule REQUEST_URI "@contains /apps/text/session/sync" \
...
+   "id:9508127,\
+    phase:1,\
+    pass,\
+    t:none,\
+    nolog,\
+    ver:'nextcloud-rule-exclusions-plugin/1.0.0',\
+    ctl:ruleRemoveTargetByTag=OWASP_CRS;ARGS:json.autosaveContent"
+
+# Text app content can be anything
+SecRule REQUEST_FILENAME "@endsWith /apps/text/public/session/push"
```
azurit, line 243: Missing ` \` at the end of line.

plugins/nextcloud-rule-exclusions-before.conf
```diff
@@ -833,4 +863,18 @@ SecRule REQUEST_FILENAME "@contains /apps/notes/notes/" \
     ctl:ruleRemoveTargetByTag=OWASP_CRS;ARGS:json.title,\
     setvar:'tx.allowed_methods=%{tx.allowed_methods} PUT DELETE'"
 
+#
+# [ Custom CSS ]
+#
+
+# Allow admin to add CSS custom CSS code
+SecRule REQUEST_FILENAME "@rx /ocs/v[0-9]\.php/apps/provisioning_api/api/v[0-9]/config/apps/theming_customcss/customcss$"
```
azurit, line 871: Missing ` \` at the end of line.

plugins/nextcloud-rule-exclusions-before.conf
EsadCetiner: the rule will not work if I add ``/`` at the end of the ``REQUEST_FILENAME``, that's how the request is made.

plugins/nextcloud-rule-exclusions-before.conf
```diff
@@ -228,7 +228,37 @@ SecRule REQUEST_URI "@contains /apps/text/session/sync" \
     nolog,\
     ver:'nextcloud-rule-exclusions-plugin/1.0.0',\
     ctl:ruleRemoveTargetByTag=OWASP_CRS;ARGS:autosaveContent"
-        
+
+# Text app content can be anything
+SecRule REQUEST_FILENAME "@endsWith /apps/text/public/session/sync"
```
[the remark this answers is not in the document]
EsadCetiner: the rule will not work if I add ``/`` at the end of the ``REQUEST_FILENAME``, that's how the request is made.

plugins/nextcloud-rule-exclusions-before.conf
```diff
@@ -228,7 +228,37 @@ SecRule REQUEST_URI "@contains /apps/text/session/sync" \
...
+   "id:9508128,\
+    phase:1,\
+    t:none,\
+    pass,\
+    nolog,\
+    ver:'nextcloud-rule-exclusions-plugin/1.0.0',\
+    ctl:ruleRemoveTargetByTag=OWASP_CRS;ARGS"
+
+# Text app content can be anything
+SecRule REQUEST_FILENAME "@endsWith /apps/text/session/push"
```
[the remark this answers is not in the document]
EsadCetiner: the rule will not work if I add ``/`` at the end of the ``REQUEST_FILENAME``, that's how the request is made.

plugins/nextcloud-rule-exclusions-before.conf
EsadCetiner: the rule will not work if I add ``/`` at the end of the ``REQUEST_FILENAME``, that's how the request is made.

plugins/nextcloud-rule-exclusions-before.conf
EsadCetiner: ``@endsWith`` should work fine here as well, I only did ``@contains`` since it was already there.

EsadCetiner:
@azurit CRS is disabled for some paramaters and locations since the input can be anything. a few examples would be having code blocks that contains SQL commands inside the text app or custom CSS code for the custom CSS app, CRS does not work well for a use case like this.

plugins/nextcloud-rule-exclusions-before.conf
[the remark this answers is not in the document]
azurit: I was talking about ` \`, not `/`.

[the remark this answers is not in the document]
azurit: `SecRule REQUEST_FILENAME "@endsWith /apps/text/public/session/sync" \`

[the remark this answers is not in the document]
EsadCetiner: done

plugins/nextcloud-rule-exclusions-before.conf
```diff
@@ -833,4 +863,18 @@ SecRule REQUEST_FILENAME "@contains /apps/notes/notes/" \
...
+
+# Allow admin to add custom CSS code
+SecRule REQUEST_FILENAME "@rx /ocs/v[0-9]\.php/apps/provisioning_api/api/v[0-9]/config/apps/theming_customcss/customcss$" \
+   "id:9508930,\
+    phase:1,\
+    pass,\
+    t:none,\
+    nolog,\
+    ver:'nextcloud-rule-exclusions-plugin/1.0.0',\
+    ctl:ruleRemoveTargetByTag=OWASP_CRS;ARGS:value"
```
azurit, line 878: What about here? Is it suppose to accept only CSS data? If so, maybe `ctl:ruleRemoveTargetByTag=attack-xss;ARGS:value` would be sufficient.

EsadCetiner: I'd rather disable CRS since SQLi and RCE rules can sometimes trigger with CSS code, disabling XSS rules should be fine at PL-1 but there will be some false positives at PL-2 and higher.

azurit (approved):

PR #27, opened 2023-06-26
Fix FP with Photos, Circles, Notes, Contacts, Files and Settings

EsadCetiner:
This pull request fixes the following
1. Minor formatting issues regarding CRS guidelines, I used the CRS rules check tool to find these issues.
2. False positives when changing settings for Photos, Contacts, Notes and Shares.
3. Fixed false positive when creating/deleting/modifying Circles.

PR #28, opened 2023-07-01
Extended text app auto sync. Notes app able to save to its settings.

TGion:
Hey Everyone,
I am pretty new to github and even less experienced with coreruleset, so any advice for improvements is very much welcome!

**Versions**
- Nextcloud 27.0.0.8
- Nginx 1.24.0 build with ModSecurity 3.0.8
- OWASP_CRS/4.0.0-rc1

**Changes**
Added a new rule ```9508341``` to allow the notes app to save to its own settings.

Extended existing rule ```9508126``` to allow get autosaving of text app working again (update of Nextcloud broke functionality).

I would really appreciate some feedback because I have no idea if this is a decent approach or totally bollocks - it works though but it doesn't mean it can't be done better. Iam fascinated by your project but is not easy for me to wrap my head around. So any tips I would appreciate very much!

### DIFF
--- a/plugins/nextcloud-rule-exclusions-before.conf
+++ b/plugins/nextcloud-rule-exclusions-before.conf
@@ -457,19 +457,8 @@ SecRule REQUEST_METHOD "@streq PROPFIND" \
 
 # We want to allow a lot of things as the user is
 # allowed to note on anything.
-SecRule REQUEST_FILENAME "@contains /index.php/apps/notes/" \
+SecRule REQUEST_FILENAME "@rx ^(?:/index.php)?/apps/notes/" \
     "id:9508340,\
-    phase:1,\
-    pass,\
-    t:none,\
-    nolog,\
-    ctl:ruleRemoveByTag=attack-injection-php,\
-    ver:'nextcloud-rule-exclusions-plugin/1.0.0',\
-    setvar:'tx.allowed_methods=%{tx.allowed_methods} PUT'"
-
-# Allow Notes App to save to its settings
-SecRule REQUEST_FILENAME "@contains /apps/notes/settings" \
-    "id:9508341,\
     phase:1,\
     pass,\
     t:none,\
--- a/plugins/nextcloud-rule-exclusions-before.conf
+++ b/plugins/nextcloud-rule-exclusions-before.conf
@@ -219,7 +219,7 @@ SecRule REQUEST_FILENAME "@contains /apps/recommendations/settings/enabled" \
     ver:'nextcloud-rule-exclusions-plugin/1.0.0',\
     setvar:'tx.allowed_methods=%{tx.allowed_methods} PUT'"
 
-# Text app autosave sync feature doesn't works 
+# Text app autosave sync feature doesn't work 
 SecRule REQUEST_URI "@contains /apps/text/session/sync" \
     "id:9508126,\
     phase:1,\
@@ -227,7 +227,7 @@ SecRule REQUEST_URI "@contains /apps/text/session/sync" \
     t:none,\
     nolog,\
     ver:'nextcloud-rule-exclusions-plugin/1.0.0',\
-    ctl:ruleRemoveTargetByTag=OWASP_CRS;ARGS:autosaveContent"
+    ctl:ruleRemoveTargetByTag=OWASP_CRS;ARGS"
 
 # Text app content can be anything
 SecRule REQUEST_FILENAME "@endsWith /apps/text/public/session/sync" \
@@ -463,8 +463,19 @@ SecRule REQUEST_FILENAME "@contains /index.php/apps/notes/" \
     t:none,\
     nolog,\
     ctl:ruleRemoveByTag=attack-injection-php,\
-    ver:'nextcloud-rule-exclusions-plugin/1.0.0'"
+    ver:'nextcloud-rule-exclusions-plugin/1.0.0',\
+    setvar:'tx.allowed_methods=%{tx.allowed_methods} PUT'"
 
+# Allow Notes App to save to its settings
+SecRule REQUEST_FILENAME "@contains /apps/notes/settings" \
+    "id:9508341,\
+    phase:1,\
+    pass,\
+    t:none,\
+    nolog,\
+    ctl:ruleRemoveByTag=attack-injection-php,\
+    ver:'nextcloud-rule-exclusions-plugin/1.0.0',\
+    setvar:'tx.allowed_methods=%{tx.allowed_methods} PUT'"
 
 #
 # [ Bookmarks ]
--- a/plugins/nextcloud-rule-exclusions-before.conf
+++ b/plugins/nextcloud-rule-exclusions-before.conf
@@ -402,6 +402,7 @@ SecRule REQUEST_METHOD "@streq PUT" \
     pass,\
     t:none,\
     nolog,\
+    ctl:ruleRemoveById=920450,\
     ver:'nextcloud-rule-exclusions-plugin/1.0.0',\
     chain"
     SecRule REQUEST_FILENAME "@contains /remote.php/dav/addressbooks/" \
--- a/plugins/nextcloud-rule-exclusions-before.conf
+++ b/plugins/nextcloud-rule-exclusions-before.conf
@@ -227,7 +227,7 @@ SecRule REQUEST_URI "@contains /apps/text/session/sync" \
     t:none,\
     nolog,\
     ver:'nextcloud-rule-exclusions-plugin/1.0.0',\
-    ctl:ruleRemoveTargetByTag=OWASP_CRS;ARGS"
+    ctl:ruleRemoveTargetByTag=OWASP_CRS;ARGS:autosaveContent"
 
 # Text app content can be anything
 SecRule REQUEST_FILENAME "@endsWith /apps/text/public/session/sync" \


plugins/nextcloud-rule-exclusions-before.conf
```diff
@@ -219,15 +219,15 @@ SecRule REQUEST_FILENAME "@contains /apps/recommendations/settings/enabled" \
...
+# Text app autosave sync feature doesn't work 
 SecRule REQUEST_URI "@contains /apps/text/session/sync" \
     "id:9508126,\
     phase:1,\
     pass,\
     t:none,\
     nolog,\
     ver:'nextcloud-rule-exclusions-plugin/1.0.0',\
-    ctl:ruleRemoveTargetByTag=OWASP_CRS;ARGS:autosaveContent"
+    ctl:ruleRemoveTargetByTag=OWASP_CRS;ARGS"
```
theseion, line 230: Can we be more restrictive here? `autosaveContent` seems to have been used once so we definitely need to keep it. What's the new name of the argument?

It would also help to document the versions the different argument names apply to.

plugins/nextcloud-rule-exclusions-before.conf
```diff
@@ -402,6 +402,7 @@ SecRule REQUEST_METHOD "@streq PUT" \
     pass,\
     t:none,\
     nolog,\
+    ctl:ruleRemoveById=920450,\
```
theseion, line 405: This would remove the rule for *all* PUT requests. Try matching against a path, like this rule does with a chain rule. Why do you need to allow restricted headers? Which ones exactly?

plugins/nextcloud-rule-exclusions-before.conf
```diff
@@ -463,8 +464,19 @@ SecRule REQUEST_FILENAME "@contains /index.php/apps/notes/" \
     t:none,\
     nolog,\
     ctl:ruleRemoveByTag=attack-injection-php,\
-    ver:'nextcloud-rule-exclusions-plugin/1.0.0'"
+    ver:'nextcloud-rule-exclusions-plugin/1.0.0',\
+    setvar:'tx.allowed_methods=%{tx.allowed_methods} PUT'"
 
+# Allow Notes App to save to its settings
+SecRule REQUEST_FILENAME "@contains /apps/notes/settings" \
```
theseion, line 471: Is this path really distinct from the on in `9508340`?

theseion (changes requested):

theseion:
Hey @TGion, thanks for the PR. I have some questions :)

theseion:
@TGion there's another PR that appears to include a super set of your changes: #27. Does that PR solve all your issues? If it does I would like to close this PR in favour of #27.

TGion:
Hey @theseion ,

I really could use some help ;)

So, I have a few questions and also some updates. I would really appreciate it if we could discuss them here:

1. After i forked the the original repo `coreruleset / nextcloud-rule-exclusions-plugin` I created a new branch for my changes and pull request. I realized when I commited the 2nd changes, I was still in this branch. Would it be more appropriate to create a seperate branch (and pull request) for every commit to seperate them or should I create one branch (and one pull request) with different commits (also seen here).
2. I have updated to the current nightly build of OWASP_CRS/4.0.0-rc1 and started with a fresh plugin configuration to get everything right. Looks like some FP arent triggered anymore. But I have a few new ones. I will comment to your questions specifically.

Thanks for your patience!

plugins/nextcloud-rule-exclusions-before.conf
TGion: Apparently. Is it possible some proxy configuration is responsible for changing the URI and is it possible (and desireable) to catch `@contains /apps/notes/settings` in any path?

plugins/nextcloud-rule-exclusions-before.conf
TGion: It was `/accept-charset/` but I did not know how to remove a restricted header from `tx.restricted_headers`.
Unfortunaly I have not observed the FP after the latest update of `OWASP_CRS/4.0.0-rc1`. Nonetheless I would appreciate the oppurtunity if you could tell me how to remove such a header from the a variable. Thanks!

theseion:
1. A single PR usually contains multiple commits that are topically related. It will not always be clear whether a change should go into a separate PR, but when in doubt, just ask. For small changes, it may sometimes be ok to put multiple changes into the same PR even if they are not related (that will also depend on the owners / maintainers / reviewers and their preferences).
So what you've done is fine.
2. RC1 is long behind us. For PRs, always base off the development branch (currently `v4.0/dev`).
If you look at https://github.com/coreruleset/coreruleset/blob/31045f09c763bb2e0bd27a45dac26950320a61fe/crs-setup.conf.example#L530 on the development branch, you will see that `Accept-Charset` is now listed as a "forbidden header" in the variable `tx.restricted_headers`.

plugins/nextcloud-rule-exclusions-before.conf
theseion: See https://github.com/coreruleset/nextcloud-rule-exclusions-plugin/blob/cecb8d59c0335b89aff9d2644a9bbd4603fe898c/plugins/nextcloud-rule-exclusions-before.conf#L52. The rules in this block are essentially copies of the rules in `crs-setup.conf`, with some modification to the default value. The rules need to be copies because of initialization order (i.e., you can't just manipulate the default value here).
If you wanted to allow `Accept-Charset` for *all* requests you would copy that rule to the same block, give it a new ID and remove `accept-charset` from the value. However, I don't think you should do that. Instead, allow `Accept-Charset` in a specific context only, just like you've done. Ideally, you would *only* make the exception for `Accept-Charset` but its not trivial to modify the variable value, so I *think* it's fine in this specific instance to allow all request headers.

plugins/nextcloud-rule-exclusions-before.conf
theseion: Its possible but unlikely that a proxy configuration manipulated the path like that. If the two rules can be combined by using `@contains /apps/notes/settings` then I would do that. Every additional rule adds a tiny overhead to request processing and duplication is bad for maintenance. But that really depends on whether there are other contexts that might be affected by the rule. Alternatively, you could change the rule to say `@rx (?:/index.php)?/apps/notes(?:/settings)?` to make it fit your requirements better (*Do not* use the expression I wrote, it's just an example. It would also match `/apps/notes` and you might not want that).

TGion: There is a rule for the notes app:

    # We want to allow a lot of things as the user is
    # allowed to note on anything.
    SecRule REQUEST_FILENAME "@contains /index.php/apps/notes/" \
        "id:9508340,\
        phase:1,\
        pass,\
        t:none,\
        nolog,\
        ctl:ruleRemoveByTag=attack-injection-php,\
        ver:'nextcloud-rule-exclusions-plugin/1.0.0'"

Adding the `REQUEST_METHOD' (Value: `PUT' )` here, still throws the FP when saving the Notes settings, because the URI is `"/apps/notes/settings"`.

I have disabled the original rule and added a new one:

    # We want to allow a lot of things as the user is
    # allowed to note on anything.
    SecRule REQUEST_FILENAME "@contains /apps/notes/" \
        "id:9508341,\
        phase:1,\
        pass,\
        t:none,\
        nolog,\
        ctl:ruleRemoveByTag=attack-injection-php,\
        setvar:'tx.allowed_methods=%{tx.allowed_methods} PUT"

Not only does saving to the settings work, but also the following actions with Notes:

- saving notes
- exotic content like scripts, commands, etc.
- assigning categories
- enable / disable favorites

Would this be a good approach to change that rule? Can you tell me, why the path `/index.php/apps/notes/` apparently works for everything expect saving to settings, but this path `/apps/notes/` works for everything.

Sorry for all those questions and thanks for your help!

TGion:
I have updated to the dev branch. So far I havent got the FP with `tx.restricted_headers`.

> 1. A single PR usually contains multiple commits that are topically related. It will not always be clear whether a change should go into a separate PR, but when in doubt, just ask. For small changes, it may sometimes be ok to put multiple changes into the same PR even if they are not related (that will also depend on the owners / maintainers / reviewers and their preferences).
>        So what you've done is fine.
> 
>     2. RC1 is long behind us. For PRs, always base off the development branch (currently `v4.0/dev`).
>        If you look at https://github.com/coreruleset/coreruleset/blob/31045f09c763bb2e0bd27a45dac26950320a61fe/crs-setup.conf.example#L530 on the development branch, you will see that `Accept-Charset` is now listed as a "forbidden header" in the variable `tx.restricted_headers`.

plugins/nextcloud-rule-exclusions-before.conf
TGion: Okay, i got the FP again. Apparently it only happens every few hours (way less than the sync - and cannot be triggered by sync):

The most elegant way (in my humble opinion) would be to remove the `Accept-Charset` from `tx.restricted_headers`.
If I would set `tx.restricted_headers` to `tx.restricted_headers=/content-encoding/ /proxy/ /lock-token/ /content-range/ /if/ /x-http-method-override/ /x-http-method/ /x-method-override/` I would override any defaults set in `crs-setup.conf`.
So with my knowledge the only way would be to allow every header in changing the rule to this:

    # Allow the data type 'text/calendar'
    SecRule REQUEST_FILENAME "@contains /remote.php/dav/calendars/" \
        "id:9508330,\
        phase:1,\
        pass,\
        t:none,\
        nolog,\
        setvar:'tx.allowed_request_content_type=%{tx.allowed_request_content_type} |text/calendar| |text/plain|',\
        setvar:'tx.restricted_headers= '"

Would this be the way to go?

TGion:
Hey @theseion ,
I am probably more work than the usual commiter and I really appreciate your time.
So my question is, can we solve all this somehow?
I have a few more FPs with the current version of CRS and NextCloud (and I think better fixes) and would like to give them back to the community.

plugins/nextcloud-rule-exclusions-before.conf
theseion: In general, yes. However, as this happens only in a very specific context I would like to be more restrictive. Do you have the log message for that hit? Then we could analyse for which request exactly the header is required. We could then write a dedicated rule for that which sets the variable to an empty string.

plugins/nextcloud-rule-exclusions-before.conf
TGion: Sweet! Thats way better. I didn't know one could do that ;)
Are those regex standarized? Can you give me a link for a good start to learn more?

I have implemented that rule and everything is working with notes EXCEPT saving the settings:

    "uri": "/apps/notes/settings"
    "message": "Method is not allowed by policy",
            "details": {
              "match": "Matched \"Operator `Within' with parameter `GET HEAD POST OPTIONS' against variable `REQUEST_METHOD' (Value: `PUT' )",
              "reference": "v0,3",
              "ruleId": "911100",
              "file": "/usr/local/etc/modsecurity/owasp-modsecurity-crs/rules/REQUEST-911-METHOD-ENFORCEMENT.conf",
              "lineNumber": "28",
              "data": "PUT",
              "severity": "2",
              "ver": "OWASP_CRS/4.0.0-rc1",
              "rev": "",
              "tags": [
                "application-multi",
                "language-multi",
                "platform-multi",
                "attack-generic",
                "paranoia-level/1",
                "OWASP_CRS",
                "capec/1000/210/272/220/274",
                "PCI/12.1"
              ],
              "maturity": "0",
              "accuracy": "0"
            }

So i have to add a regex to `SecRule REQUEST_FILENAME "@rx ^(?/index.php)?/apps/notes/` to also catch `settings`?
So we could avoid using `contains`?
Something like you mentioned above: `@rx (?:/index.php)?/apps/notes(?:/settings)?`

plugins/nextcloud-rule-exclusions-before.conf
TGion: I have disabled the rule and hope to trigger that FP again. Ill get back to you.
Thanks again for your time!

plugins/nextcloud-rule-exclusions-before.conf
theseion: In general, our regular expressions are Re2 compatible (https://github.com/google/re2/wiki/Syntax).

The rule should have matched. `/app/notes/settings` clearly begins with `/apps/notes/`. But I see that I made an error in the rule. It should be 
```
"@rx ^(?:/index.php)?/apps/notes/"
```
instead of
```
"@rx ^(?/index.php)?/apps/notes/"
```

theseion: Sounds good. But drop the `(?:/settings)?`. For one, it introduces a double slash. Also, it is superfluous, the matches are identical with and without that piece.

TGion: Sorry for my late reply @theseion 
And thanks for your help. It works as suggested.
How do we proceed from here?  Do I have to make another pull request with the updated rule or can I change my current one?

Thanks again for your patience. I really appreciate it!

TGion: Oh, and also we should change the id to: `9508340`
I dont know why it ended up with the id `9508341`

theseion: Just update this PR.

theseion:
@TGion please resolve all unresolved conversations in this PR. I'll do a fresh review then.

TGion:
@theseion 
I think we can dismiss this whole pull request.:

1. Extended text app auto sync. Notes app able to save to its settings. This one is obsolete with https://github.com/coreruleset/nextcloud-rule-exclusions-plugin/pull/27
2. Calendar: Restricted header detected: /accept-charset/ fixed. I can't reproduce the FP.

Sorry for taking up your time with this.
I will consider it a learning exercise though.

What do you think?

theseion:
Sure, I'll close this PR. Thanks for the contribution anyway. I hope it wasn't your last one :)

Issue #2, opened 2022-04-29
Variable expansion not working?


mathsen:
### Describe the bug
I have a nextcloud instance and just switched from the "old" exclusions to the new plugin system.
During the tests I reckognized a problem:
[Rule 9508130](https://github.com/coreruleset/nextcloud-rule-exclusions-plugin/blob/65015038c617f58d6fe63a6c33e983f6dda7b21e/plugins/nextcloud-rule-exclusions-before.conf#L202) should expand tx.allowed_methods:
````
setvar:'tx.allowed_methods=%{tx.allowed_methods} PUT PATCH CHECKOUT COPY DELETE LOCK MERGE MKACTIVITY MKCOL MOVE PROPFIND PROPPATCH SEARCH UNLOCK REPORT TRACE jsonp'"`
````
tx.allowed_methods is the default value (GET HEAD POST OPTIONS), so I don't set it - but somehow it is not expanded here.
I receive the following block in my log:

````
---xbaiLTEY---A--
[29/Apr/2022:17:06:53 +0200] XXXX
---xbaiLTEY---B--
GET /index.php/avatar/xxx/384 HTTP/1.1
[...]
Host: XXX
User-Agent: Mozilla/5.0 (Android) Nextcloud-android/3.19.1
[...]


---xbaiLTEY---F--
HTTP/1.1 403
Date: Fri, 29 Apr 2022 15:06:53 GMT
Content-Length: 146
Content-Type: text/html
Connection: keep-alive

---xbaiLTEY---H--
[...]

PR #29, opened 2023-07-31
Address FP in NC 27 file manager - details - versioning and system tags

TGion:
**Versions**
- Nextcloud 27.0.0.8
- Nginx 1.24.0 build with ModSecurity 3.0.8
- OWASP_CRS/4.0.0-nightly (updated)

**Issue**
Accessing file details in the file manager throws FPs for the following URLs:

     /remote.php/dav/systemtags-relations/files/
     /remote.php/dav/systemtags/
     /remote.php/dav/versions/_USERNAME_/versions/

**Reason**

    TX:content_type' (Value: `|text/plain|`

is not allowed.

I can provide detailed logs if needed.
Thanks for your consideration.

### DIFF
--- a/plugins/nextcloud-rule-exclusions-before.conf
+++ b/plugins/nextcloud-rule-exclusions-before.conf
@@ -135,6 +135,39 @@ SecRule REQUEST_FILENAME "@contains /remote.php/dav/files/" \
     ver:'nextcloud-rule-exclusions-plugin/1.0.0',\
     setvar:'tx.allowed_request_content_type=%{tx.allowed_request_content_type} |text/plain| |text/vcard|'"
 
+# File manager: versioning
+# Allow the data type 'text/plain'
+SecRule REQUEST_FILENAME "@contains /remote.php/dav/versions/" \
+    "id:9508111,\
+    phase:1,\
+    pass,\
+    t:none,\
+    nolog,\
+    ver:'nextcloud-rule-exclusions-plugin/1.0.0',\
+    setvar:'tx.allowed_request_content_type=%{tx.allowed_request_content_type} |text/plain|'"
+
+# File manager: system tags
+# Allow the data type 'text/plain'
+SecRule REQUEST_FILENAME "@contains /remote.php/dav/systemtags/" \
+    "id:9508112,\
+    phase:1,\
+    pass,\
+    t:none,\
+    nolog,\
+    ver:'nextcloud-rule-exclusions-plugin/1.0.0',\
+    setvar:'tx.allowed_request_content_type=%{tx.allowed_request_content_type} |text/plain|'" 
+
+# File manager: system tags
+# Allow the data type 'text/plain'
+SecRule REQUEST_FILENAME "@contains /remote.php/dav/systemtags-relations/" \
+    "id:9508113,\
+    phase:1,\
+    pass,\
+    t:none,\
+    nolog,\
+    ver:'nextcloud-rule-exclusions-plugin/1.0.0',\
+    setvar:'tx.allowed_request_content_type=%{tx.allowed_request_content_type} |text/plain|'" 
+
 # Allow the data type 'application/octet-stream'
 SecRule REQUEST_METHOD "@pm PUT MOVE" \
     "id:9508115,\
--- a/plugins/nextcloud-rule-exclusions-before.conf
+++ b/plugins/nextcloud-rule-exclusions-before.conf
@@ -137,36 +137,29 @@ SecRule REQUEST_FILENAME "@contains /remote.php/dav/files/" \
 
 # File manager: versioning
 # Allow the data type 'text/plain'
+# Since the content is actually XML, we switch on the XML parser
 SecRule REQUEST_FILENAME "@contains /remote.php/dav/versions/" \
     "id:9508111,\
     phase:1,\
     pass,\
     t:none,\
     nolog,\
+    ctl:requestBodyProcessor=XML,\
     ver:'nextcloud-rule-exclusions-plugin/1.0.0',\
     setvar:'tx.allowed_request_content_type=%{tx.allowed_request_content_type} |text/plain|'"
 
 # File manager: system tags
 # Allow the data type 'text/plain'
-SecRule REQUEST_FILENAME "@contains /remote.php/dav/systemtags/" \
+# Since the content is actually XML, we switch on the XML parser
+SecRule REQUEST_FILENAME "@rx /remote.php/dav/systemtags(?:-relations)/" \
     "id:9508112,\
     phase:1,\
     pass,\
     t:none,\
     nolog,\
+    ctl:requestBodyProcessor=XML,\
     ver:'nextcloud-rule-exclusions-plugin/1.0.0',\
-    setvar:'tx.allowed_request_content_type=%{tx.allowed_request_content_type} |text/plain|'" 
-
-# File manager: system tags
-# Allow the data type 'text/plain'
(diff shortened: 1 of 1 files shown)


RedXanadu:
@TGion Out of interest, do you know what kind of data is transferred in the body of those requests that have a content type of `text/plain`?

The use of `text/plain` opens up a specific bypass opportunity: the WAF doesn't know how to process the request body in a meaningful way. For that reason, `text/plain` is not permitted any more by default. Allowing `text/plain` potentially allows any arbitrary request bodies to pass undetected, containing any kind of malicious payload.

TGion:
Hey @RedXanadu ,
thanks for you reply. I think I understand your concerns, but I am afraid my level of knowledge and skill is limited. Digging through the source code is not a task I feel capable at the moment.

Nonetheless, here is what I found out about one of the actual requests:

**Request Header**

        PROPFIND /remote.php/dav/systemtags-relations/files/19323/ HTTP/2
        Host: cloud.gion.io
        User-Agent: Mozilla/5.0 (X11; Linux x86_64; rv:109.0) Gecko/20100101 Firefox/115.0
        Accept: text/plain,application/xml
        Accept-Language: de,en-US;q=0.7,en;q=0.3
        Accept-Encoding: gzip, deflate, br
        Content-Type: text/plain;charset=UTF-8
        Content-Length: 220
        requesttoken: ***
        Depth: 1
        Origin: https://cloud.gion.io
        Sec-Fetch-Dest: empty
        Sec-Fetch-Mode: cors
        Sec-Fetch-Site: same-origin
        Connection: keep-alive
        Cookie: oc_sessionPassphrase=***
        Pragma: no-cache
        Cache-Control: no-cache
        TE: trailers

**Request Body**

      <?xml version="1.0"?>
      <d:propfind  xmlns:d="DAV:" xmlns:oc="http://owncloud.org/ns">
	      <d:prop>
		      <oc:id />
		      <oc:display-name />
		      <oc:user-visible />
		      <oc:user-assignable />
		      <oc:can-assign />
	      </d:prop>
      </d:propfind>

Out of curiosity: If we would allow, for example `text/xml` for the content-type, the request would still fail, because `text/plain` is requested in the request header, correct?

theseion:
>Out of curiosity: If we would allow, for example text/xml for the content-type, the request would still fail, because text/plain is requested in the request header, correct?

Yes.

For exclusion rules it is acceptable to allow `text/plain` in specific contexts, just like we had discussed in your other PR for request methods.

RedXanadu:
Sorry for the late reply!

I would be tempted to write a rule exclusion that has a `ctl` action to switch on the XML parser, as this `plain` content appears to actually be XML under the covers.

Simply allowing through `text/plain` is a simpler option and should work, but if you go down that road then I would **strongly** recommend making it very clear in the comments that such a rule exclusion will mean that request bodies will not be inspected, for the reasons mentioned above.

TGion:
No problem @RedXanadu and thanks for your reply!
That was actually going to be my next question: Could we run the XML parser on a `text/plain` request ;)
Thanks for answering before I even asked!
If I would rate the possible solutions, the list would go like:

1. Change the request in the NC source code to `text/xml` to properly match the content. I don't really wanna go down that rabbit hole right now.
2. Create a rule to accept `text/plain` but also switch on the XML parser.
3. Leave it like suggested and add the appropriate comments to the rule.

I would prefer option #2. Could you help me to extend the rule to switch on the XML parser?

PS: @theseion Thanks for your answer, I really appreciate it.

theseion:
@TGion to enable the XML body processor you would add a `ctl` statement:
```
...
ctl:requestBodyProcessor=XML
...
```

TGion:
Sorry for my late reply and thanks for your support.
I have checked all 3 requests and they indeed all contain XML in their request body. So I switched on the XML parser in all three rules like you guys suggested. The rules are working and no FP is generated.

1. How can I check the XML parser is actually working? If the XML parser would fail for some reason (not XML content, suspicious XML code, etc.) the rule would simply fail and therefore the FP would be triggered, correct?
2. I have (tried) to combine the previous rules `9508112` and `9508112` because of their similarity in the `REQUEST_FILENAME`:
- `/remote.php/dav/systemtags/`
- `/remote.php/dav/systemtags-relations/`

I have come up with this regex: `REQUEST_FILENAME "@rx /remote.php/dav/systemtags(?:-relations)/"`
The `systemtags-relations` works, `systemtags` not. Can you help me correct the regex so the rule works for both?

Again, thanks a lot for your help!

theseion:
Your regex is missing a`?`:
```
REQUEST_FILENAME "@rx /remote.php/dav/systemtags(?:-relations)?/"
```

TGion:
Thanks @theseion,
I have tested the new regex and it catches both URLs as intended.
So Iam pretty happy with the solution.

theseion:
> How can I check the XML parser is actually working? If the XML parser would fail for some reason (not XML content, suspicious XML code, etc.) the rule would simply fail and therefore the FP would be triggered, correct?

No. If the parser fails then `REQBODY_ERROR` will be set and you should check 
its value: https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual-%28v2.x%29#reqbody_error.

theseion:
`REQBODY_ERROR` is checked in rule `200002` if you are using the ModSecurity recommended rules: https://github.com/SpiderLabs/ModSecurity/blob/v3/master/modsecurity.conf-recommended#L75.

TGion:
Thanks for the insigt @theseion , it's really fascinating what there still is to discover for me.
Iam using the recommend rules and I can also see the rule is disabled several times in the plugin.

Anything else I can do here?

plugins/nextcloud-rule-exclusions-before.conf
```diff
@@ -135,6 +135,32 @@ SecRule REQUEST_FILENAME "@contains /remote.php/dav/files/" \
     ver:'nextcloud-rule-exclusions-plugin/1.0.0',\
     setvar:'tx.allowed_request_content_type=%{tx.allowed_request_content_type} |text/plain| |text/vcard|'"
 
+# File manager: versioning
+# Allow the data type 'text/plain'
+# Since the content is actually XML, we switch on the XML parser
+SecRule REQUEST_FILENAME "@contains /remote.php/dav/versions/" \
```
theseion, line 141:
```suggestion
SecRule REQUEST_FILENAME "@beginsWith /remote.php/dav/versions/" \
```

plugins/nextcloud-rule-exclusions-before.conf
```diff
@@ -135,6 +135,32 @@ SecRule REQUEST_FILENAME "@contains /remote.php/dav/files/" \
...
+    t:none,\
+    nolog,\
+    ctl:requestBodyProcessor=XML,\
+    ver:'nextcloud-rule-exclusions-plugin/1.0.0',\
+    setvar:'tx.allowed_request_content_type=%{tx.allowed_request_content_type} |text/plain|'"
+
+# File manager: system tags
+# Allow the data type 'text/plain'
+# Since the content is actually XML, we switch on the XML parser
+SecRule REQUEST_FILENAME "@rx /remote.php/dav/systemtags(?:-relations)?/" \
```
theseion, line 154:
```suggestion
SecRule REQUEST_FILENAME "@rx ^/remote.php/dav/systemtags(?:-relations)?/" \
```

theseion (approved):
Always anchor searches when possible

TGion:
@theseion Thanks for your help here.
Is there anything for me left to do to finish the merge?

theseion (approved):

theseion:
Thanks @TGion!

PR #30, opened 2023-08-09
Add support for Nextcloud Mail

EsadCetiner:
Fixes false positives with the Nextcloud Mail client

### DIFF
--- a/plugins/nextcloud-rule-exclusions-before.conf
+++ b/plugins/nextcloud-rule-exclusions-before.conf
@@ -732,7 +732,7 @@ SecRule REQUEST_FILENAME "@endsWith /apps/theming/background/custom" \
     setvar:'tx.allowed_methods=%{tx.allowed_methods} DELETE'"
 
 # When deleting an admin from privacy page
-SecRule REQUEST_FILENAME "@rx /apps/privacy/api/admins/[0-9]*$" \
+SecRule REQUEST_FILENAME "@rx /apps/privacy/api/admins/[0-9]+$" \
     "id:9508618,\
     phase:1,\
     pass,\
@@ -1043,7 +1043,7 @@ SecRule REQUEST_FILENAME "@rx /apps/files/api/v[0-9]/views/trashbin/(?:sorting_d
     setvar:'tx.allowed_methods=%{tx.allowed_methods} PUT'"
 
 # Rejecting a file ownership transfer
-SecRule REQUEST_FILENAME "@rx /ocs/v[0-9]\.php/apps/files/api/v[0-9]/transferownership/[0-9]*$" \
+SecRule REQUEST_FILENAME "@rx /ocs/v[0-9]\.php/apps/files/api/v[0-9]/transferownership/[0-9]+$" \
     "id:9508963,\
     phase:1,\
     pass,\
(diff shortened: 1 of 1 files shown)


plugins/nextcloud-rule-exclusions-before.conf
```diff
@@ -1078,4 +1078,232 @@ SecRule REQUEST_FILENAME "@rx /ocs/v[0-9]\.php/apps/files/api/v[0-9]/transferown
...
+#
+
+# Some email addresses end in .com or .inc
+SecRule REQUEST_FILENAME "@rx /apps/mail/api/autoconfig/ispdb/[^/]+(?:com|inc)$" \
+    "id:9508970,\
+    phase:1,\
+    pass,\
+    t:none,\
+    nolog,\
+    ctl:ruleRemoveById=920440,\
```
azurit, line 1092: This and also rule below (`9508971`) should be fixed by similar way as `allowed_methods`, see rules 9508100 and for example `9508122`.

plugins/nextcloud-rule-exclusions-before.conf
```diff
@@ -1078,4 +1078,232 @@ SecRule REQUEST_FILENAME "@rx /ocs/v[0-9]\.php/apps/files/api/v[0-9]/transferown
     ver:'nextcloud-rule-exclusions-plugin/1.0.0',\
     setvar:'tx.allowed_methods=%{tx.allowed_methods} DELETE'"
 
+#
+# [ Nextcloud Mail ]
+#
+
+# Some email addresses end in .com or .inc
+SecRule REQUEST_FILENAME "@rx /apps/mail/api/autoconfig/ispdb/[^/]+(?:com|inc)$" \
```
azurit, line 1086: I would add also a dot at the end of regexp:
`/apps/mail/api/autoconfig/ispdb/[^/]+\.(?:com|inc)$`

plugins/nextcloud-rule-exclusions-before.conf
```diff
@@ -1078,4 +1078,232 @@ SecRule REQUEST_FILENAME "@rx /ocs/v[0-9]\.php/apps/files/api/v[0-9]/transferown
...
+SecRule REQUEST_FILENAME "@rx /apps/mail/api/accounts/[0-9]+/draft$" \
+    "id:9508976,\
+    phase:1,\
+    pass,\
+    t:none,\
+    nolog,\
+    ctl:ruleRemoveTargetByTag=OWASP_CRS;ARGS:json.editorBody,\
+    ctl:ruleRemoveTargetByTag=OWASP_CRS;ARGS:json.body,\
+    ver:'nextcloud-rule-exclusions-plugin/1.0.0',\
+    ver:'nextcloud-rule-exclusions-plugin/1.0.0'"
```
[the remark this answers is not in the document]
EsadCetiner: should've been fixed by https://github.com/coreruleset/nextcloud-rule-exclusions-plugin/pull/30/commits/75d3b72f0d14c33dc50ab19deb1bb259936136a0 can you verify?

plugins/nextcloud-rule-exclusions-before.conf
EsadCetiner: I'm not sure how that would work, I'm not adding a restricted file extensions like in ``9508122`` but removing? it should respect any changes made by the end user since it only matches file extensions ending in ``(com|inc)``

azurit: I just realised it probably cannot be done as we need to remove something from a list, not add it. Any ideas?

azurit: Ok, let's keep it like it is now.

plugins/nextcloud-rule-exclusions-before.conf
EsadCetiner: Thanks, I missed that

azurit:
I just noticed that marker `END-NEXTCLOUD-ADMIN` was never used so let's remove it - pls delete line `SecMarker "END-NEXTCLOUD-ADMIN"` at the end of the file. Thanks.

azurit:
Merging, thank you!

PR #32, opened 2023-08-28
URL content triggers FP from rule 933120 when syncing

TGion:
**Versions**

    FreeBSD 13.2-RELEASE-p2
    nginx version: nginx/1.24.0 
    CRS v4.0/dev 
    Floccus Extension 4.19.1 (Firefox Browser)
    
**Issue**

Certain URLs trigger a FP when the content matches `./rules/php-config-directives.data`.
In this particular case, *engine* is matched.
The affected URL is https://engineer.kodekloud.com/?utm_source=Linkedin&utm_medium=social&utm_campaign=Introducing+KKE+2.0+-+Mumshad+Video&utm_content=video

    "request": {
      "method": "PUT",
      "http_version": 2,
      "uri": "/index.php/apps/bookmarks/public/rest/v2/bookmark/767",
      "headers": {
        "sec-fetch-mode": "cors",
        "sec-fetch-dest": "empty",
        "authorization": "Basic ***",
        "host": "cloud.gion.io",
        "sec-fetch-site": "same-origin",
        "user-agent": "Mozilla/5.0 (X11; Linux x86_64; rv:109.0) Gecko/20100101 Firefox/115.0",
        "accept": "*/*",
        "content-length": "238",
        "te": "trailers",
        "accept-language": "de,en-US;q=0.7,en;q=0.3",
        "accept-encoding": "gzip, deflate, br",
        "content-type": "application/json",
        "origin": "moz-extension://67ef67e2-7fb3-4831-b211-332cbd2d6c36"
      }

    --- snip ---

    "producer": {
          "modsecurity": "ModSecurity v3.0.8 (FreeBSD)",
          "connector": "ModSecurity-nginx v1.0.3",
          "secrules_engine": "Enabled",
          "components": [
            "OWASP_CRS/4.0.0-rc1\""
          ]
        },
        "messages": [
          {
            "message": "PHP Injection Attack: Configuration Directive Found",
            "details": {
              "match": "Matched \"Operator `Pm' with parameter `=' against variable `MATCHED_VARS:ARGS:json.url' (Value: `https:/engineer.kodekloud.com/?utm_source=Linkedin&utm_medium=social&utm_campaign=Introducing+KKE+2. (35 characters omitted)' )",
              "reference": "o7,6v9,136t:normalisePatho10,6v11,56t:normalisePatho41,1v584,135",
              "ruleId": "933120",
              "file": "/usr/local/etc/modsecurity/owasp-modsecurity-crs/rules/REQUEST-933-APPLICATION-ATTACK-PHP.conf",
              "lineNumber": "112",
              "data": "Matched Data: engine found within MATCHED_VARS:ARGS:json.url: https:/engineer.kodekloud.com/?utm_source=Linkedin&utm_medium=social&utm_campaign=Introducing+KKE+2.0+-+Mumshad+Video&utm_content=video",
              "severity": "2",
              "ver": "OWASP_CRS/4.0.0-rc1",
              "rev": "",
              "tags": [
                "application-multi",
                "language-php",
                "platform-multi",
                "attack-injection-php",
                "paranoia-level/1",
                "OWASP_CRS",
                "capec/1000/152/242"
              ],
              "maturity": "0",
              "accuracy": "0"
            }
          },

**Solution**

I suggest we disable rule `933120` because URLs may include any of the keywords in `php-config-directives.data`.
Also made the exclusion little bit more strict by changing `@contains` to `@beginsWith` in the REQUEST_FILENAME.

If there is a more secure and elegant way, I am happy for you suggestion. Otherwise, thanks for your consideration!

### DIFF
--- a/plugins/nextcloud-rule-exclusions-before.conf
+++ b/plugins/nextcloud-rule-exclusions-before.conf
@@ -508,13 +508,14 @@ SecRule REQUEST_FILENAME "@contains /index.php/apps/notes/" \
 #
 
 # Allow urls in data.
-SecRule REQUEST_FILENAME "@contains /index.php/apps/bookmarks/" \
+SecRule REQUEST_FILENAME "@beginsWith /index.php/apps/bookmarks/" \
     "id:9508350,\
     phase:1,\
     pass,\
     t:none,\
     nolog,\
     ver:'nextcloud-rule-exclusions-plugin/1.0.0',\
+    ctl:ruleRemoveById=933120,\
     ctl:ruleRemoveById=931130"
 
 


TGion:
@RedXanadu Thanks for clearing that up.
Looks like the mentioned small post processing was not done:

# Small post-processing is needed to remove numbers coming from a column,
# and `*` chars (e.g. `pdo.dsn.*`).
# Also removed single words like `engine`, `extension`, `from` and `precision`, to prevent FP.

Yes, I can open a pull request. But who will decide on those words? Do you have any suggestions on how to approach that?

- Clearly those words mentioned earlier should be removed (*engine* being the one which triggered the FP for me)
- How to decide on any more single words which should be removed

TGion:
@theseion Thanks for clearing that up.
So, how do we proceed? Personally I can't evaluate the impact of having the rule disabled for the bookmarks endpoint. I would need your expertise on how to proceed further.
Please tell me if there is anything I can contribute to.
Thanks!

theseion:
@dune73 Would appreciate your opinion here.

dune73:
I think the assessment, namely from @RedXanadu is correct. In fact we are facing a very similar situation to the 93315x and 93316x rule family overhaul merged in #3273 (don't bother to read the entire thread @TGion). The point is there is PHP stuff that also appears in the English language. In 3273, we covered this via a pm-rule and a separate regex-rule for those English terms. Broadly speaking.

Given these are "only" configuration directives and it's not that many. I would probably skip them for CRS in 933120 and make this transparent as a comment to the rule. If you think this is too risky, then a regex variant of this rule is due, either as stricter sibling at higher PL, or as a related rule also at higher PL.

theseion:
Thanks. Personally, I'd go with an additional rule at PL2. Otherwise we can't claim that we fixed the BB issue.

dune73:
That's true. I agree.

I'm leaning on stricter sibling as 933121.

I am not entirely sure about all uses of these keywords in php.ini and friends. 

So are we looking at `keyword\s*=` or is it more complex than this?

theseion:
Here's what I think we should do:

1. New rule 933121 at PL 2
3. Move all single word entries (no `.` or `_`) to this new rule
4. Use regular expression to match:
  - `keyword=\s*=`
  - `ini_set\s*\(\s*keyword` (see https://www.php.net/manual/en/function.ini-set.php)

@TGion If you wanted to attempt this, you would create a new file for the new rule in the `regex-assembly` directory. This will will be used by crs-toolchain to generate the regular expression.

dune73:
Not sure about `keyword=\s*=`. Should not this be `keyword\s*=`?

theseion:
Of course, thanks. I've fixed it in the comment.

TGion:
Thanks for your input everyone.
@theseion Yes, I would like to give it a shot. I will probably come back with (many) questions ;)
Until then, I wish everyone a good time!

theseion:
Great! You might want to consider joining our Slack channel (see https://github.com/coreruleset/coreruleset), where you could also DM me with questions, might be easier.

theseion:
Closing due to inactivity. @TGion feel free to reopen.

PR #33, opened 2023-09-12
Remove Duplicate rules, fix FPs and Tighten a few rules

EsadCetiner:
This PR fixes the following
1. Marks capturing group regex as non capturing
2. Tighten rule exclusion rules to be more precise
3. Fix FP when disabling dyslexia font
4. Fix FP when viewing the trashbin
5. Fix FP with Sharing files/folders
6. Remove duplicate rules

### DIFF
--- a/plugins/nextcloud-rule-exclusions-before.conf
+++ b/plugins/nextcloud-rule-exclusions-before.conf
@@ -236,7 +236,7 @@ SecRule REQUEST_FILENAME "@contains /remote.php/dav/comments" \
     setvar:'tx.allowed_request_content_type=%{tx.allowed_request_content_type} |text/plain|'"
 
 # When trying to change the "Show recommendations" option in file settings
-SecRule REQUEST_FILENAME "@contains /apps/recommendations/settings/enabled" \
+SecRule REQUEST_FILENAME "@endsWith /apps/recommendations/settings/enabled" \
     "id:9508124,\
     phase:1,\
     pass,\
@@ -582,7 +582,7 @@ SecRule REQUEST_FILENAME "@endsWith /settings/users" \
 # Personal: Security
 
 # Fixes deletion and allow/disallow of filesystem access for auth tokens
-SecRule REQUEST_FILENAME "@contains /settings/personal/authtokens/" \
+SecRule REQUEST_FILENAME "@rx /settings/personal/authtokens/[0-9]+$" \
     "id:9508601,\
     phase:1,\
     pass,\
@@ -602,7 +602,7 @@ SecRule REQUEST_FILENAME "@contains /settings/api/personal/webauthn/registration
     setvar:'tx.allowed_methods=%{tx.allowed_methods} DELETE'"
 
 # While going through the first run setup wizard
-SecRule REQUEST_FILENAME "@contains /apps/firstrunwizard/wizard" \
+SecRule REQUEST_FILENAME "@endsWith /apps/firstrunwizard/wizard" \
     "id:9508607,\
     phase:1,\
     pass,\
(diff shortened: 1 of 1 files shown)


plugins/nextcloud-rule-exclusions-before.conf
```diff
@@ -126,13 +126,20 @@ SecRule REQUEST_METHOD "@pm PROPFIND PUT" \
         ctl:ruleRemoveById=921110"
 
 # Allow the data type 'text/vcard'
+# Allow characters like /../ in files.
+# Allow all kind of filetypes.
```
theseion, line 130:
```suggestion
# Allow all filetypes.
```

plugins/nextcloud-rule-exclusions-before.conf
```diff
@@ -126,13 +126,20 @@ SecRule REQUEST_METHOD "@pm PROPFIND PUT" \
         ctl:ruleRemoveById=921110"
 
 # Allow the data type 'text/vcard'
+# Allow characters like /../ in files.
```
theseion, line 129: Path traversal is a common security issue. Can you explain why this is necessary? I know you didn't put that exclusion here, but it would be helpful, if you know the reason.

```suggestion
# Allow path traversal characters like /../ in file paths.
```

plugins/nextcloud-rule-exclusions-before.conf
```diff
@@ -226,17 +218,20 @@ SecRule REQUEST_FILENAME "@contains /apps/text/session/create" \
     setvar:'tx.allowed_methods=%{tx.allowed_methods} PUT'"
 
 # FP while trying to open text files in file manager
-SecRule REQUEST_FILENAME "@contains /remote.php/dav/comments" \
+# Adding/editing comments to files
+# The content of comments could be anything
+SecRule REQUEST_FILENAME "@rx /remote.php/dav/comments/files/[0-9]+(?:/[0-9]+)?$" \
```
theseion, line 223:
```suggestion
SecRule REQUEST_FILENAME "@rx /remote\.php/dav/comments/files(?:/[0-9]+)+$" \
```

theseion (changes requested):

plugins/nextcloud-rule-exclusions-before.conf
EsadCetiner: That's intentional.
the rule is meant to match both:
```
/remote.php/dav/comments/files/123/45
/remote.php/dav/comments/files/123
```
I'll fix the unescaped ``.`` character.

theseion: My version matches both. It's equivalent to what you wrote initially, just shorter.

EsadCetiner: you're right, I just tested with regexr. Fixed.

plugins/nextcloud-rule-exclusions-before.conf
EsadCetiner: I'm not sure the reasoning behind it and I don't want to break things for other users by overtightening rules. I'm on libmodsecurity 3 and iirc excluding by rule id ranges doesn't work so maybe it's safe to remove?

theseion: If you don't know, than that's fine. Leave it as it is.

EsadCetiner: I'll leave as is then, I'll fix it in the future if I do know.

theseion (approved):
Thanks once more for you awesome contributions to this plugin!

Issue #34, opened 2023-09-17
Erroneously launching XML parser


tysino:
Hi,
rule 9508112 switches on XML parser for all calls to `/remote\.php/dav/systemtags(?:-relations)?`, like so:

```
# File manager: system tags
# Allow the data type 'text/plain'
# Since the content is actually XML, we switch on the XML parser
SecRule REQUEST_FILENAME "@rx ^/remote\.php/dav/systemtags(?:-relations)?/" \
    "id:9508112,\
    phase:1,\
    pass,\
    t:none,\
    nolog,\
    ver:'nextcloud-rule-exclusions-plugin/1.0.0',\
    ctl:requestBodyProcessor=XML,\
    setvar:'tx.allowed_request_content_type=%{tx.allowed_request_content_type} |text/plain|'"
```

However, this should only be done if the content-type is actually `text/plain`. Nextcloud also calls this path with content-type `application/json` (when creating and assigning tags), in which case the XML parser obviously fails. This then results in an internal server error.

I fixed this issue on my machine using an additional *-after.conf plugin file with a chained rule, where the XML parser is only loaded for `text/plain`:
```
# fix XML parsing error where nextcloud actually uses JSON

SecRuleRemoveById 9508112

# File manager: system tags
# Allow the data type 'text/plain'
# if content-type is 'text/plain', the content is actually XML. so we switch on the XML parser
SecRule REQUEST_FILENAME "@rx ^/remote\.php/dav/systemtags(?:-relations)?/" \
    "id:999999998,\
    phase:1,\
    pass,\
    t:none,\
    nolog,\
    ver:'nextcloud-rule-exclusions-plugin/1.0.0',\
    setvar:'tx.allowed_request_content_type=%{tx.allowed_request_content_type} |text/plain|'"
    SecRule REQUEST_HEADERS:Content-Type "text/plain" \
        "id:999999999,\
         phase:1,\
         t:none,\
         nolog,\
         ctl:requestBodyProcessor=XML"
```

I don't know if this is a good solution or if it breaks other rules, so any comment would be appreciated.

EsadCetiner:
@tysino I've opened a pull request to fix the XML parser failing. I'd appreciate it if you could test it and see if it resolves your issue since I don't use the tagging feature in Nextcloud. However, from my testing, it seems to resolve the issue.
Please see: https://github.com/coreruleset/nextcloud-rule-exclusions-plugin/pull/35

PR #35, opened 2023-09-18
Fix XML parser failing

EsadCetiner:
Fixes issue https://github.com/coreruleset/nextcloud-rule-exclusions-plugin/issues/34


theseion (approved):
LGTM

PR #36, opened 2023-09-19
fix FP for photos app (albums, sharedalbums, places)

mhastu:
Allow text/plain and switch on XML parser for /remote.php/dav/photos/\<username\>/albums|sharedalbums|places.

Please review the username regex ([allowed characters](https://docs.nextcloud.com/server/latest/admin_manual/configuration_user/user_configuration.html#creating-a-new-user)). I'm not sure if the urlencoded '@' character is necessary.

### DIFF
--- a/plugins/nextcloud-rule-exclusions-before.conf
+++ b/plugins/nextcloud-rule-exclusions-before.conf
@@ -1227,7 +1227,7 @@ SecRule REQUEST_FILENAME "@rx /apps/photos/api/v[0-9]+/preview/[0-9]+$" \
 # Photos: Albums, Shared Albums, Places
 # Allow the data type 'text/plain'
 # Since the content is actually XML, we switch on the XML parser
-SecRule REQUEST_FILENAME "@rx ^/remote\.php/dav/photos/(?:[a-zA-Z0-9\-\_\. \@]|\%40)*/(?:(?:shared)?albums|places)/" \
+SecRule REQUEST_FILENAME "@rx /remote\.php/dav/photos/[^/]+/(?:albums|sharedalbums|places)(?:/[^/]+)?/$" \
     "id:9508952,\
     phase:1,\
     pass,\
--- a/plugins/nextcloud-rule-exclusions-before.conf
+++ b/plugins/nextcloud-rule-exclusions-before.conf
@@ -1224,6 +1224,22 @@ SecRule REQUEST_FILENAME "@rx /apps/photos/api/v[0-9]+/preview/[0-9]+$" \
         "t:none,\
         ctl:ruleRemoveById=950100"
 
+# Photos: Albums, Shared Albums, Places
+# Allow the data type 'text/plain'
+# Since the content is actually XML, we switch on the XML parser
+SecRule REQUEST_FILENAME "@rx ^/remote\.php/dav/photos/(?:[a-zA-Z0-9\-\_\. \@]|\%40)*/(?:(?:shared)?albums|places)/" \
+    "id:9508952,\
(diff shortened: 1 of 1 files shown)


plugins/nextcloud-rule-exclusions-before.conf
```diff
@@ -1062,6 +1062,22 @@ SecRule REQUEST_FILENAME "@rx /apps/photos/api/v[0-9]/config/config/(?:croppedLa
     ver:'nextcloud-rule-exclusions-plugin/1.0.0',\
     setvar:'tx.allowed_methods=%{tx.allowed_methods} PUT'"
 
+# Photos: Albums, Shared Albums, Places
+# Allow the data type 'text/plain'
+# Since the content is actually XML, we switch on the XML parser
+SecRule REQUEST_FILENAME "@rx ^/remote\.php/dav/photos/(?:[a-zA-Z0-9\-\_\. \@]|\%40)*/(?:(?:shared)?albums|places)/" \
```
theseion, line 1068: This should work. Please test.
```suggestion
SecRule REQUEST_FILENAME "@rx ^/remote\.php/dav/photos/[^/]+/(?:(?:shared)?albums|places)$" \
```

theseion (changes requested):

theseion:
@EsadCetiner would you take a look at this PR too please?

plugins/nextcloud-rule-exclusions-before.conf
EsadCetiner: I agree with @theseion to allow all characters within the username field even if Nextcloud only allows a certain set of characters. I don't think there will be much gain (if any) in limiting it, plus it looks a little cleaner that way. Also, the suggested regex although stricter does have FPs, this one ``/remote\.php/dav/photos/[^/]+/(?:albums|sharedalbums|places)(?:/[^/]+)?/$
`` from my testing it works well and accommodates users that may have installed nextcloud under ``/nextcloud/`` or ``/my-nextcloud/`` for example.

EsadCetiner (changes requested):

EsadCetiner (approved):
LGTM

theseion (approved):

plugins/nextcloud-rule-exclusions-before.conf
```diff
@@ -1062,6 +1062,22 @@ SecRule REQUEST_FILENAME "@rx /apps/photos/api/v[0-9]/config/config/(?:croppedLa
     ver:'nextcloud-rule-exclusions-plugin/1.0.0',\
     setvar:'tx.allowed_methods=%{tx.allowed_methods} PUT'"
 
+# Photos: Albums, Shared Albums, Places
+# Allow the data type 'text/plain'
+# Since the content is actually XML, we switch on the XML parser
+SecRule REQUEST_FILENAME "@rx /remote\.php/dav/photos/[^/]+/(?:albums|sharedalbums|places)(?:/[^/]+)?/$" \
+    "id:9508951,\
```
EsadCetiner: I just realized that rule id ``9508951`` is already in use, you'll have to pick another rule id like ``9508952`` to avoid a conflict.
https://github.com/coreruleset/nextcloud-rule-exclusions-plugin/blob/327425915e1ca48accaaba6b8ca06fd4f13cfc82/plugins/nextcloud-rule-exclusions-before.conf#L1216
```suggestion
    "id:9508952,\
```

theseion: Just fixed :)

theseion:
Thank you @mhastu!

PR #39, opened 2023-09-29
Fix Recognize, Passwords, Cookie, Text Editor, Search, and deleting mailbox FPs

EsadCetiner:
This PR fixes the following
1. Adds support for Nextcloud Recognize
2. Fix FP when changing passwords
3. Fix FP when changing sorting direction in Files
4. Fix FP when deleting a mailbox
5. Fix FP with cookies
6. Fix FP with text editor
7. Fix another search FP
8. Removes some duplicate rules (and also fixes another FP)
9. Fix FP when setting up a new Nextcloud Server

### DIFF
--- a/plugins/nextcloud-rule-exclusions-before.conf
+++ b/plugins/nextcloud-rule-exclusions-before.conf
@@ -744,7 +744,7 @@ SecRule REQUEST_FILENAME "@endsWith /index.php" \
     nolog,\
     ver:'nextcloud-rule-exclusions-plugin/1.0.0',\
     chain"
-    SecRule ARGS:dbtype "@rx ^(mysql|pgsql)$" \
+    SecRule ARGS:dbtype "@rx ^(?:my|pg)sql$" \
         "t:none,\
         ctl:ruleRemoveTargetById=932236;ARGS:dbtype,\
         ctl:ruleRemoveTargetByTag=OWASP_CRS;ARGS:adminpass,\
--- a/plugins/nextcloud-rule-exclusions-before.conf
+++ b/plugins/nextcloud-rule-exclusions-before.conf
@@ -289,15 +289,21 @@ SecRule REQUEST_FILENAME "@endsWith /apps/text/session/push" \
 
 # Viewing files within the trashbin
 # Content is XML so the XML parser is switched on
-SecRule REQUEST_FILENAME "@rx /remote\.php/dav/trashbin/[^/]+/trash/$" \
+SecRule REQUEST_FILENAME "@rx /remote\.php/dav/trashbin/[^/]+/trash/" \
     "id:9508170,\
     phase:1,\
     pass,\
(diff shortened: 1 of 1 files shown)


plugins/nextcloud-rule-exclusions-before.conf
```diff
@@ -275,27 +268,23 @@ SecRule REQUEST_FILENAME "@endsWith /apps/text/public/session/push" \
...
     t:none,\
     nolog,\
-    ctl:requestBodyProcessor=XML,\
     ver:'nextcloud-rule-exclusions-plugin/1.0.0',\
-    setvar:'tx.allowed_request_content_type=%{tx.allowed_request_content_type} |text/plain|'"
+    chain"
+    SecRule REQUEST_HEADERS:Content-Type "@rx ^text/plain(?:\;charset=UTF-8)?$" \
+        "t:none,\
+        chain"
+        SecRule REQUEST_METHOD "@streq PROPFIND" \
```
theseion, line 284: Does this check also work for older versions?

plugins/nextcloud-rule-exclusions-before.conf
```diff
@@ -720,6 +707,31 @@ SecRule REQUEST_FILENAME "@endsWith /settings/users" \
...
+    phase:1,\
+    pass,\
+    t:none,\
+    nolog,\
+    ver:'nextcloud-rule-exclusions-plugin/1.0.0',\
+    ctl:ruleRemoveTargetById=942450;REQUEST_COOKIES:oc_sessionPassphrase,\
+    ctl:ruleRemoveTargetById=942210;REQUEST_COOKIES:oc_sessionPassphrase,\
+    ctl:ruleRemoveTargetById=942450;REQUEST_COOKIES:nc_session_id,\
+    ctl:ruleRemoveTargetById=942450;REQUEST_COOKIES:nc_token,\
+    ctl:ruleRemoveTargetById=942450;REQUEST_COOKIES:ocddjoynjqgd"
```
theseion, line 733: Is this value really stable??

theseion (changes requested):

plugins/nextcloud-rule-exclusions-before.conf
EsadCetiner: hm... it looks to be unique to each Nextcloud instance. It's a PL-2 FP so it's not a big deal but still annoying.

plugins/nextcloud-rule-exclusions-before.conf
EsadCetiner: Yes, the FP was introduced in Nextcloud 27 Hub 5 (Previous release) and it works there.
I encountered another FP while double checking this and added another commit for it

theseion (approved):

theseion:
Thanks for also taking care of clean up!

PR #40, opened 2023-10-10
fix: don't inspect REQUEST_FILENAME of uploads

### LINKED ISSUE #38: Allow arbitrary filenames
I noticed that the regex of rule 933210 (PHP injection) is also applied on REQUEST_FILENAME. Is this intended behavior? I.e. do these kind of attacks also work by trying to sync special filenames or should we allow arbitrary filenames? Granted, using quotes in filenames is a bad idea, but if you receive a bunch of files and just want to save it in your cloud it is unexpected that suddenly the syncing fails, just because of the filename.

The relevant error.log:

``
[Mon Sep 25 09:47:56.108156 2023] [security2:error] [pid 132240] [client 0.0.0.0:0] [client 0.0.0.0] ModSecurity: Warning. Pattern match "(?:(?:\\\\(|\\\\[)[a-zA-Z0-9_.$\\"'\\\\[\\\\](){}/*\\\\s]+(?:\\\\)|\\\\])[0-9_.$\\"'\\\\[\\\\](){}/*\\\\s]*\\\\([a-zA-Z0-9_.$\\"'\\\\[\\\\](){}/*\\\\s].*\\\\)|\\\\([\\\\s]*string[\\\\s]*\\\\)[\\\\s]*(?:\\"|'))" at REQUEST_FILENAME. [file "/etc/modsecurity/crs/rules/REQUEST-933-APPLICATION-ATTACK-PHP.conf"] [line "503"] [id "933210"] [msg "PHP Injection Attack: Variable Function Call Found"] [data "Matched Data: (text)' (text) found within REQUEST_FILENAME: /remote.php/dav/files/text - 'text (text)' (text) - text.ext"] [severity "CRITICAL"] [ver "OWASP_CRS/3.3.5"] [tag "application-multi"] [tag "language-php"] [tag "platform-multi"] [tag "attack-injection-php"] [tag "paranoia-level/1"] [tag "OWASP_CRS"] [tag "capec/1000/152/242"] [hostname "example.url"] [uri "/remote.php/dav/files/text - 'text (text)' (text) - text.ext"] [unique_id "id"]
``

theseion:
- Ignore PHP injections on file names during uploads
- Remove duplicate rule

Fixes #38

### DIFF
--- a/plugins/nextcloud-rule-exclusions-before.conf
+++ b/plugins/nextcloud-rule-exclusions-before.conf
@@ -129,6 +129,7 @@ SecRule REQUEST_METHOD "@pm PROPFIND PUT" \
 # Allow path traversal characters like /../ in file paths.
 # Allow all filetypes.
 # Allow source code.
+# Allow arbitrary filenames.
 SecRule REQUEST_FILENAME "@contains /remote.php/dav/files/" \
     "id:9508110,\
     phase:1,\
@@ -140,6 +141,7 @@ SecRule REQUEST_FILENAME "@contains /remote.php/dav/files/" \
     ctl:ruleRemoveById=951000-951999,\
     ctl:ruleRemoveById=953100-953130,\
     ctl:ruleRemoveById=920440,\
+    ctl:ruleRemoveTargetById=933210;REQUEST_FILENAME,\
     setvar:'tx.allowed_request_content_type=%{tx.allowed_request_content_type} |text/plain| |text/vcard|'"
 
 # File manager: versioning
@@ -317,19 +319,6 @@ SecRule REQUEST_FILENAME "@rx /remote\.php/dav/files/[^/]+/$" \
     ver:'nextcloud-rule-exclusions-plugin/1.0.0',\
     chain"
     SecRule RESPONSE_STATUS "@rx ^50(?:3|2)$" \
-        "t:none,\
-        ctl:ruleRemoveById=950100"
-
-# Syncing files with Nextcloud desktop app
-SecRule REQUEST_FILENAME "@contains /remote.php/dav/files/" \
-    "id:9508174,\
-    phase:3,\
-    pass,\
-    t:none,\
-    nolog,\
-    ver:'nextcloud-rule-exclusions-plugin/1.0.0',\
-    chain"
-    SecRule RESPONSE_STATUS "@streq 500" \
         "t:none,\
         ctl:ruleRemoveById=950100"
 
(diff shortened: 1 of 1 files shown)


theseion:
@EsadCetiner I'd appreciate it if you would take a look at this PR.

EsadCetiner (approved):
LGTM

EsadCetiner (changes requested):
Actually, this could be removed as well.

```diff
-# Syncing files with Nextcloud desktop app
-SecRule REQUEST_FILENAME "@rx /remote\.php/dav/files/[^/]+/$" \
-    "id:9508173,\
-    phase:3,\
-    pass,\
-    t:none,\
-    nolog,\
-    ver:'nextcloud-rule-exclusions-plugin/1.0.0',\
-    chain"
-    SecRule RESPONSE_STATUS "@rx ^50(?:3|2)$" \
-        "t:none,\
-        ctl:ruleRemoveById=950100"
```

theseion:
I don't see which rule would cover the same thing as `9508173`. Could you point me to it?

plugins/nextcloud-rule-exclusions-before.conf
```diff
@@ -320,19 +322,6 @@ SecRule REQUEST_FILENAME "@rx /remote\.php/dav/files/[^/]+/$" \
...
-    phase:3,\
-    pass,\
-    t:none,\
-    nolog,\
-    ver:'nextcloud-rule-exclusions-plugin/1.0.0',\
-    chain"
-    SecRule RESPONSE_STATUS "@streq 500" \
-        "t:none,\
-        ctl:ruleRemoveById=950100"
-
```
EsadCetiner, line 335: My mistake, I thought this was ``ctl:ruleRemoveById=950000-950999,\`` when it's actually ``ctl:ruleRemoveById=951000-951999,\`` https://github.com/coreruleset/nextcloud-rule-exclusions-plugin/pull/40/files#diff-b927017cb1fee736fcfe55bdef8b2ca3f4966a9896ea93d3e1e6f29487303be0R141

In that case, removing rules ``9508173`` and or ``9508174`` will introduce a regression. ``9508173`` is meant to match ``/remote.php/dav/files/username/`` and ``9508174`` is meant to match ``/remote.php/dav/files/path/to/example.txt``.

You could bring back rule ``9508174`` and modify ``RESPONSE_STATUS`` to match ``^50(?:3|2|0)$`` and remove ``9508173`` or leave as is.

theseion: Well spotted, thanks. I've pushed a change that should work for both rules.

azurit (approved):

plugins/nextcloud-rule-exclusions-before.conf
EsadCetiner: You also need to modify the chained rule to accept status codes of 500 ```^50(?:3|2|0)$```

theseion: Ups, missed that. Thanks.

EsadCetiner (approved):

PR #41, opened 2023-10-17
refact: move NC files exclusion to file manager

EsadCetiner:
Moves rule exclusions under the catagory Nextcloud Files to File Manager, they are both the same thing. I haven't touched any of the rule IDs.


azurit:
Are those rules really related to a File Manager feature?

EsadCetiner:
Yes, Nextcloud Files is Nextcloud's file manager app

Issue #5, opened 2022-11-04
False positives


feanorknd:
Hi there, lots of 921110 errors...

``
ModSecurity: Warning. Matched "Operator `Rx' with parameter `(?:get|post|head|options|connect|put|delete|trace|track|patch|propfind|propatch|mkcol|copy|move|lock|unlock)\s+(?:\/|\w)[^\s]*(?:\s+http\/\d|[\r\n])' against variable `REQUEST_BODY' (Value: `<?xml version="1.0" ?>\x0a<d:propfind xmlns:d="DAV:">\x0a  <d:prop>\x0a    <d:getetag/>\x0a  </d:pro (23 characters omitted)' ) [file "/etc/modsecurity/owasp/coreruleset/rules/REQUEST-921-PROTOCOL-ATTACK.conf"] [line "33"] [id "921110"] [rev ""] [msg "HTTP Request Smuggling Attack"] [data "Matched Data: propfind xmlns:d="dav:">\x0a found within REQUEST_BODY: <?xml version="1.0" ?>\x0a<d:propfind xmlns:d="dav:">\x0a  <d:prop>\x0a    <d:getetag/>\x0a  </d:prop>\x0a</d:propfind>\x0a"] [severity "2"] [ver "OWASP_CRS/3.3.2"] [maturity "0"] [accuracy "0"] [tag "application-multi"] [tag "language-multi"] [tag "platform-multi"] [tag "attack-protocol"] [tag "paranoia-level/1"] [tag "OWASP_CRS"] [tag "capec/1000/210/272/220/33"] [hostname "xxxxxxxxx"] [uri "/remote.php/dav/files/a.xxxxxxxxx/"] [unique_id "166756463217.934518"] [ref "o26,25v844,105t:urlDecodeUni,t:htmlEntityDecode,t:lowercase"]
``

just detecting:

```
---fIkF7ECV---C--
<?xml version="1.0" ?>
**<d:propfind xmlns:d="DAV:">**
  <d:prop>
    <d:getetag/>
  </d:prop>
</d:propfind>
```

Maybe could be good to include an exception for this.

Thanks.

theseion:
Same issue as https://github.com/coreruleset/coreruleset/issues/2973, but for the new plugin.

EsadCetiner:
[...]

PR #42, opened 2023-10-26
Fix: various file upload FPs

EsadCetiner:
This PR fixes most FPs relating to file uploads, there still may be one or two that I missed.

I also noticed some exclusions commented out, should I document them in the readme as well since it's a bit hidden?
https://github.com/EsadCetiner/nextcloud-rule-exclusions-plugin/blob/7e2c4064e050febc06cbdb093fb80867ece88fb2/plugins/nextcloud-rule-exclusions-before.conf#L23

closes https://github.com/coreruleset/nextcloud-rule-exclusions-plugin/issues/5 and https://github.com/coreruleset/coreruleset/issues/2973

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,6 +63,6 @@ After the plugin is enabled, Nextcloud should work without problems caused by CR
 
 ## License
 
-Copyright (c) 2023 OWASP ModSecurity Core Rule Set project. All rights reserved.
+Copyright (c) 2022-2023 OWASP ModSecurity Core Rule Set project. All rights reserved.
 
 The OWASP ModSecurity Core Rule Set and its official plugins are distributed under Apache Software License (ASL) version 2. Please see the enclosed LICENSE file for full details.
--- a/README.md
+++ b/README.md
@@ -24,7 +24,16 @@ SecRule REQUEST_URI "@endsWith /index.php/apps/files/ajax/upload.php" \
 ctl:requestBodyLimit is not supported in libmodsecurity3, Nginx users can increase max upload size
 by using the following:
 
-```location /index.php/apps/files/ajax/upload.php { modsecurity_rules 'SecRequestBodyLimit 1073741824'; }```
+```
+location /index.php/apps/files/ajax/upload.php { modsecurity_rules 'SecRequestBodyLimit 1073741824'; }
+```
+
+Apache libmodsecurity3 Example:
+```
+ <location "/index.php/apps/files/ajax/upload.php">
+    modsecurity_rules 'SecRequestBodyLimit 1073741824'
+</location>
+```
 
 ## Relaxing file upload restrictions
 
@@ -54,6 +63,13 @@ Apache modsecurity2 Example:
 ```
 <location "/remote.php/dav/files/">
    SecRequestBodyNoFilesLimit 144384
+</location>
+```
+
+Apache libmodsecurity3 Example:
+```
+<location "/remote.php/dav/files/">
+   modsecurity_rules 'SecRequestBodyNoFilesLimit 144384'
 </location>
 ```
 
--- a/README.md
+++ b/README.md
@@ -8,12 +8,61 @@ This plugin contains rule exclusions for [Nextcloud](https://nextcloud.com/), a 
 
 For full and up to date instructions for the different available plugin installation methods, refer to [How to Install a Plugin](https://coreruleset.org/docs/concepts/plugins/#how-to-install-a-plugin) in the official CRS documentation.
 
+## Increasing max upload size
+
+Large uploads can be modified with SecRequestBodyLimit. Or they can be more controlled by using the following:
+
+```
+SecRule REQUEST_URI "@endsWith /index.php/apps/files/ajax/upload.php" \
+    "id:9508610,\
+    phase:1,\
+    t:none,\
+    nolog,\
+    ctl:requestBodyLimit=1073741824"
+```
+
+ctl:requestBodyLimit is not supported in libmodsecurity3, Nginx users can increase max upload size
+by using the following:
+
+```location /index.php/apps/files/ajax/upload.php { modsecurity_rules 'SecRequestBodyLimit 1073741824'; }```
+
+## Relaxing file upload restrictions
+
+To relax upload restrictions for only the php files that need it, you put something like this in crs-setup.conf:
+
+```
+SecRule REQUEST_FILENAME "@rx /(?:remote\.php|index\.php)/" \
+    "id:9508600,\
+    phase:2,\
+    t:none,\
+    nolog,\
+    pass,\
+    setvar:'tx.restricted_extensions=.bak/ .config/ .conf/'"
+```
+
+## Increasing max request body size
+
+The Nextcloud desktop client occasionally sends large request bodies not containing any uploaded files.
+ModSecurity will block request bodies larger than 131KB, adjusting SecRequestBodyNoFilesLimit to 141KB works for all scenarios tested.
+
+Nginx libmodsecurity3 Example:
+```
+location /remote.php/dav/files/ { modsecurity_rules 'SecRequestBodyNoFilesLimit 144384'; }
+```
+
+Apache modsecurity2 Example:
+```
+<location "/remote.php/dav/files/">
+   SecRequestBodyNoFilesLimit 144384
+</location>
+```
+
 ## Testing
 
 After the plugin is enabled, Nextcloud should work without problems caused by CRS (for example, false positives while blocking requests). If problems still occur then please file a new issue on [GitHub](https://github.com/coreruleset/nextcloud-rule-exclusions-plugin). (Note that high paranoia level deployments may require additional tuning beyond this plugin.)
 
 ## License
 
-Copyright (c) 2022 OWASP ModSecurity Core Rule Set project. All rights reserved.
+Copyright (c) 2023 OWASP ModSecurity Core Rule Set project. All rights reserved.
 
 The OWASP ModSecurity Core Rule Set and its official plugins are distributed under Apache Software License (ASL) version 2. Please see the enclosed LICENSE file for full details.
--- a/plugins/nextcloud-rule-exclusions-before.conf
+++ b/plugins/nextcloud-rule-exclusions-before.conf
@@ -1,6 +1,6 @@
 # ------------------------------------------------------------------------
 # OWASP ModSecurity Core Rule Set Plugin
-# Copyright (c) 2021-2022 Core Rule Set project. All rights reserved.
+# Copyright (c) 2021-2023 Core Rule Set project. All rights reserved.
 #
 # The OWASP ModSecurity Core Rule Set plugins are distributed under
 # Apache Software License (ASL) version 2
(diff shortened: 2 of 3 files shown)


plugins/nextcloud-rule-exclusions-before.conf
```diff
@@ -27,28 +27,40 @@ SecRule TX:nextcloud-rule-exclusions-plugin_enabled "@eq 0" "id:9508099,phase:1,
...
 #
 # Please note, that `ctl:requestBodyLimit` action does not work
 # in libmodsecurity3!
+#
+# The Nextcloud desktop client occasionally sends large request bodies not containing any uploaded files.
+# ModSecurity will block request bodies larger than 131KB, changing the SecRequestBodyNoFilesLimit
+# directive to a higher value for the file path/remote.php/dav/files/ fixes this issue.
+#
+# Nginx libmodsecurity3 Example:
+# location /remote.php/dav/files/ { modsecurity_rules 'SecRequestBodyNoFilesLimit 141662'; }
```
theseion, line 55: Where does this number come from?

plugins/nextcloud-rule-exclusions-before.conf
```diff
@@ -27,28 +27,40 @@ SecRule TX:nextcloud-rule-exclusions-plugin_enabled "@eq 0" "id:9508099,phase:1,
...
+#     "id:9508610,\
+#     phase:1,\
+#     t:none,\
+#     nolog,\
+#     ctl:requestBodyLimit=1073741824"
 #
 # Please note, that `ctl:requestBodyLimit` action does not work
 # in libmodsecurity3!
+#
+# The Nextcloud desktop client occasionally sends large request bodies not containing any uploaded files.
```
theseion, line 50: I think this new section should live in the readme instead. Maybe reference the readme in the paragraph on lines 37, 38 for this particular case.

plugins/nextcloud-rule-exclusions-before.conf
```diff
@@ -142,6 +154,11 @@ SecRule REQUEST_FILENAME "@contains /remote.php/dav/files/" \
     ctl:ruleRemoveById=953100-953130,\
     ctl:ruleRemoveById=920440,\
     ctl:ruleRemoveTargetById=933210;REQUEST_FILENAME,\
+    ctl:ruleRemoveTargetById=920220;REQUEST_URI,\
```
theseion, line 157: Please explain these exclusions in the rule comment as was done for the others. Please order by ID.

plugins/nextcloud-rule-exclusions-before.conf
```diff
@@ -326,6 +343,39 @@ SecRule REQUEST_FILENAME "@contains /remote.php/dav/files/" \
...
+# This FP is usually triggered when uploading via the mobile app
+SecRule REQUEST_FILENAME "@rx /remote\.php/dav/uploads/[^/]+/[0-9]+/[0-9]+$" \
+    "id:9508174,\
+    phase:1,\
+    pass,\
+    t:none,\
+    nolog,\
+    ver:'nextcloud-rule-exclusions-plugin/1.0.0',\
+    ctl:ruleRemoveById=920341,\
+    ctl:ruleRemoveTargetById=944200;REQUEST_BODY,\
```
theseion, line 356: Please order by ctl action, then ID.

plugins/nextcloud-rule-exclusions-before.conf
```diff
@@ -326,6 +343,39 @@ SecRule REQUEST_FILENAME "@contains /remote.php/dav/files/" \
...
+# Uploading files in bulk
+SecRule REQUEST_FILENAME "@endsWith /remote.php/dav/bulk" \
+    "id:9508175,\
+    phase:1,\
+    pass,\
+    t:none,\
+    nolog,\
+    ver:'nextcloud-rule-exclusions-plugin/1.0.0',\
+    ctl:ruleRemoveById=921422,\
+    ctl:ruleRemoveTargetById=944250;REQUEST_BODY,\
```
theseion, line 373: Pleas order by ctl action, then ID.

theseion (changes requested):

plugins/nextcloud-rule-exclusions-before.conf
EsadCetiner: I'm not sure what you mean, do you mean doing something like this?
```
    ctl:ruleRemoveTargetById=944200;REQUEST_BODY,\
    ctl:ruleRemoveById=920341"
```

plugins/nextcloud-rule-exclusions-before.conf
EsadCetiner: it came from my own testing, I haven't seen a request body larger than 141kb.

plugins/nextcloud-rule-exclusions-before.conf
EsadCetiner: I think the comments here explain it well enough, should I add "Uploading files"?

EsadCetiner: also, you want me to order the rule IDs highest to lowest or lowest to highest?

plugins/nextcloud-rule-exclusions-before.conf
theseion: Fair enough, just add that to the comment. How did you compute the number?
1024 * 141 = 144384.
144662 / 1024 = 138.341796875.

plugins/nextcloud-rule-exclusions-before.conf
theseion: I went to look at the rule numbers (e.g., 920220) and it wasn't immediately clear why the rules would need to be excluded. I suppose in this case it would be because Nextcloud performs invalid URL encoding?
The reason I'd like the comments is so when we revisit this rule at some point in the future we can decide whether it still makes sense for a given ID to be excluded. If we don't know why it's there we end up guessing. This sort of thing has lead to many problems in the past for this project.

Sort IDs from lowest to highest.

plugins/nextcloud-rule-exclusions-before.conf
theseion: No, sorry. I just wanted you to sort by ID, but by ctl action first, so that all rules with the same ctl action are grouped together, not the other way around (in this case that doesn't matter though, I had just copied my comment :) ).

plugins/nextcloud-rule-exclusions-before.conf
EsadCetiner: Sorry, I meant bytes not bits, 141KB. I found that number from trial and error.

plugins/nextcloud-rule-exclusions-before.conf
EsadCetiner: gotcha, lots of rules in the plugin aren't ordered in that manner so i'll change those as well

plugins/nextcloud-rule-exclusions-before.conf
EsadCetiner: I get what you mean now, I'll try and improve the comments for that rule

plugins/nextcloud-rule-exclusions-before.conf
theseion: Ok. I suggest rounding up to 144384, then you can just say that 141KB worked for all tested scenarios.

plugins/nextcloud-rule-exclusions-before.conf
EsadCetiner: Agreed, I've documented them in readme so they aren't hidden.

plugins/nextcloud-rule-exclusions-before.conf
EsadCetiner: I can't reproduce and find the cause of 920220 FP, maybe the FP doesn't exist in newer Nextcloud version, should I remove it and re-add in the future when I can find the cause? it's only a warning so it's not a big deal.

I can reproduce the request body FPs like 944120 but I'm not sure what comment to put for it, 944120 is triggered by ``https://gamedownloads.rockstargames.com/mtl/js/20220405bcbd/build/launcher.fc6801e3b29d7eae931e.js/*! for license information please see launcher.fc6801e3b29d7eae``, should I put something along the lines of "allow upload of files with funny content"?

theseion: Yes, remove 920220.

How about "Allow uploads of files containing code (Java, JavaScript, etc.)"?

EsadCetiner: yeah that sounds good to me

theseion (approved):

theseion:
Ping @azurit.

README.md
```diff
@@ -8,12 +8,61 @@ This plugin contains rule exclusions for [Nextcloud](https://nextcloud.com/), a
...
+```
+SecRule REQUEST_URI "@endsWith /index.php/apps/files/ajax/upload.php" \
+    "id:9508610,\
+    phase:1,\
+    t:none,\
+    nolog,\
+    ctl:requestBodyLimit=1073741824"
+```
+
+ctl:requestBodyLimit is not supported in libmodsecurity3, Nginx users can increase max upload size
```
azurit, line 24: What about Apache users?

README.md
```diff
@@ -8,12 +8,61 @@ This plugin contains rule exclusions for [Nextcloud](https://nextcloud.com/), a
...
+```
+
 ## Testing
 
 After the plugin is enabled, Nextcloud should work without problems caused by CRS (for example, false positives while blocking requests). If problems still occur then please file a new issue on [GitHub](https://github.com/coreruleset/nextcloud-rule-exclusions-plugin). (Note that high paranoia level deployments may require additional tuning beyond this plugin.)
 
 ## License
 
-Copyright (c) 2022 OWASP ModSecurity Core Rule Set project. All rights reserved.
+Copyright (c) 2023 OWASP ModSecurity Core Rule Set project. All rights reserved.
```
azurit, line 66: Should be 2022-2023

plugins/nextcloud-rule-exclusions-before.conf
```diff
@@ -27,28 +27,41 @@ SecRule TX:nextcloud-rule-exclusions-plugin_enabled "@eq 0" "id:9508099,phase:1,
...
-#    t:none,\
-#    nolog,\
-#    ctl:requestBodyLimit=1073741824"
+#     "id:9508610,\
+#     phase:1,\
+#     t:none,\
+#     nolog,\
+#     ctl:requestBodyLimit=1073741824"
+#
+# ctl:requestBodyLimit is not supported in libmodsecurity3, Nginx users can increase max upload size
```
azurit, line 47: Example for Apache is missing also here.

README.md
theseion: They're covered by `requestBodyLimit`.

plugins/nextcloud-rule-exclusions-before.conf
theseion: As above, httpd is already covered by `requestBodyLimit`.

README.md
azurit: But it says that this isn't supported by libmodsecurity3.

theseion: Oh, sorry. My bad. You were referring to httpd on libmodsecurity 3. I always forget that's a thing 😅

EsadCetiner: it doesn't look like libmodsecurity3 apache is still supported, no commits in 2.5 years and it's in beta?

EsadCetiner: I've added an example nonetheless

azurit (approved):

azurit:
Thank you, merging!

PR #47, opened 2024-01-07
fix: file uploads FP from Windows client

### LINKED ISSUE #46: FP when uploading files via desktop client
Uploading files from the (windows) desktop client produces false positives with the log:

`ModSecurity: Warning. String match within "/accept-charset/ /content-encoding/ /proxy/ /lock-token/ /content-range/ /if/" at TX:header_name_if. [file "/etc/modsecurity/crs/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf"] [line "1128"] [id "920450"] [msg "HTTP header is restricted by policy (/if/)"] [data "Restricted header detected: /if/"] [severity "CRITICAL"] [ver "OWASP_CRS/3.3.5"] [tag "application-multi"] [tag "language-multi"] [tag "platform-multi"] [tag "attack-protocol"] [tag "paranoia-level/1"] [tag "OWASP_CRS"] [tag "capec/1000/210/272"] [tag "PCI/12.1"] [hostname "example.org"] [uri "/remote.php/dav/uploads/user/4195284221/.file"]`

I managed to fix it on my setup by removing `/if` form the restricted header list for requests uris of this form, with the following rule:
```
SecRule REQUEST_FILENAME "@rx /remote\.php/dav/uploads/[^/]+/[0-9]+/.file$" \
    "id:9999999999,\
    phase:1,\
    pass,\
    t:none,\
    nolog,\
    setvar:'tx.restricted_headers=/accept-charset/ /content-encoding/ /proxy/ /lock-token/ /content-range/'"
```

I think that *removing* `/if` from the current value of `tx.restricted_headers` might be more future-proof, but I don't know how to do that, hence I didn't open a pull-request.

Desktop client version: 3.11.0 (might be relevant, since this issue did not occur before)

EsadCetiner:
closes #46

### DIFF
--- a/plugins/nextcloud-rule-exclusions-before.conf
+++ b/plugins/nextcloud-rule-exclusions-before.conf
@@ -387,6 +387,19 @@ SecRule REQUEST_FILENAME "@endsWith /remote.php/dav/bulk" \
     ctl:ruleRemoveTargetById=944240;REQUEST_BODY,\
     ctl:ruleRemoveTargetById=944250;REQUEST_BODY"
 
+# Uploading files via Windows desktop client
+SecRule REQUEST_FILENAME "@rx /remote\.php/dav/uploads/[^/]+/[0-9]+/[a-z0-9]+$" \
+    "id:9508176,\
+    phase:1,\
+    pass,\
+    t:none,\
+    nolog,\
+    ver:'nextcloud-rule-exclusions-plugin/1.0.0',\
+    chain"
+    SecRule &REQUEST_HEADERS:if "@eq 1" \
+        "t:none,\
+        ctl:ruleRemoveById=920450"
+
 #
 # [ Searchengine ]
 #
--- a/plugins/nextcloud-rule-exclusions-before.conf
+++ b/plugins/nextcloud-rule-exclusions-before.conf
@@ -388,7 +388,7 @@ SecRule REQUEST_FILENAME "@endsWith /remote.php/dav/bulk" \
     ctl:ruleRemoveTargetById=944250;REQUEST_BODY"
 
 # Uploading files via Windows desktop client
-SecRule REQUEST_FILENAME "@rx /remote\.php/dav/uploads/[^/]+/[0-9]+/[a-z0-9\.]+$" \
+SecRule REQUEST_FILENAME "@rx /remote\.php/dav/uploads/[^/]+/[0-9]+/\.file$" \
     "id:9508176,\
     phase:1,\
     pass,\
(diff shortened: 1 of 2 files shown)


mhastu (changes requested):
The regex for the request filename won't work. All request filenames I have encountered in my FPs matched `/remote\.php/dav/uploads/[^/]+/[0-9]+/.file$`. Keeping the restrictive `.file` at the end is working so far.

EsadCetiner:
@mhastu the uri path you provided in your issue should match, is a different uri path not matching?
https://regexr.com/7q35l

or is the if header is sent twice?

I'll change the regex to ``\.file$`` at the end if your sure that all requests end with that, it just seems a bit strange

mhastu:
Apologies, I misread the regex and didn't notice that `[a-z0+9\.]` will match `.file`. And you're right, there's no need to be overly restrictive with the ending. But in that case, I would handle it like the username and change `[a-z0-9\.]` to `[^/]`, so it can match any filename.

EsadCetiner:
but have you actually seen requests to ``remote.php/dav/uploads/`` ending in a filename? I'm happy to change the regex to check if it's ending in ``.file`` or to a negated set ``[^/]+`` if it's a filename.

mhastu:
No, all the requests of FPs on my system ended in `.file`.

EsadCetiner:
ok, thank you!

If you encounter another FP where the request uri doesn't end in ``.file`` then let me know and we can improve the regex.

tests/regression/nextcloud-rule-exclusions-plugin/9508176.yaml
```diff
@@ -0,0 +1,23 @@
+---
+meta:
+  author: "Esad Cetiner"
+  description: "Nextcloud Rule Exclusions Plugin"
+  enabled: true
+  name: 9508176.yaml
+tests:
+  - test_title: 9508176-1
+    desc: Allow the if header when uploading files from Window's desktop client
```
theseion, line 9:
```suggestion
    desc: Allow the if header when uploading files from Windows desktop client
```

theseion (changes requested):

plugins/nextcloud-rule-exclusions-before.conf
```diff
@@ -387,6 +387,19 @@ SecRule REQUEST_FILENAME "@endsWith /remote.php/dav/bulk" \
     ctl:ruleRemoveTargetById=944240;REQUEST_BODY,\
     ctl:ruleRemoveTargetById=944250;REQUEST_BODY"
 
+# Uploading files via Windows desktop client
+SecRule REQUEST_FILENAME "@rx /remote\.php/dav/uploads/[^/]+/[0-9]+/\.file$" \
```
theseion, line 391: In the PR for the Android client you used `[a-z0-9]`, I believe. Is that necessary here too?

EsadCetiner: a false positive wasn't reported within the a-z character range so it should be fine

theseion (approved):

PR #48, opened 2024-01-18
fix: xml parser failure when restoring file to old version

EsadCetiner:
this fixes a regression when restoring a file to an older version by only using the xml parser for text/plain content type

### DIFF
--- a/plugins/nextcloud-rule-exclusions-before.conf
+++ b/plugins/nextcloud-rule-exclusions-before.conf
@@ -182,8 +182,11 @@ SecRule REQUEST_FILENAME "@beginsWith /remote.php/dav/versions/" \
     t:none,\
     nolog,\
     ver:'nextcloud-rule-exclusions-plugin/1.0.0',\
-    ctl:requestBodyProcessor=XML,\
-    setvar:'tx.allowed_request_content_type=%{tx.allowed_request_content_type} |text/plain|'"
+    chain"
+    SecRule REQUEST_HEADERS:Content-Type "@rx ^text/plain(?:\;charset=UTF-8)?$" \
+        "t:none,\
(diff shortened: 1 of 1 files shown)


plugins/nextcloud-rule-exclusions-before.conf
```diff
@@ -182,8 +182,11 @@ SecRule REQUEST_FILENAME "@beginsWith /remote.php/dav/versions/" \
     t:none,\
     nolog,\
     ver:'nextcloud-rule-exclusions-plugin/1.0.0',\
-    ctl:requestBodyProcessor=XML,\
-    setvar:'tx.allowed_request_content_type=%{tx.allowed_request_content_type} |text/plain|'"
+    chain"
+    SecRule REQUEST_HEADERS:Content-Type "@rx ^text/plain(?:\;charset=UTF-8)?$" \
```
theseion, line 186: This expression is very restrictive. In Asia, for example, the charset may not be utf-8.
```suggestion
    SecRule REQUEST_HEADERS:Content-Type "@rx (?i)^text/plain(?:\;charset=utf-8)?$" \
```

theseion (changes requested):

plugins/nextcloud-rule-exclusions-before.conf
EsadCetiner: I'm not sure how making the regex case insensitive would help, if you wanted to support charsets other than UTF-8 then wouldn't it make more sense to use ``@beginsWith text/plain`` or specify the charset within the regex?

theseion: Yes, absolutely. The question was unrelated to the change I suggested to the regex.

theseion (approved):

PR #53, opened 2024-02-11
Fix: False positive in deck application

mivek:
I created this PR to handle false positive when using the deck application in nextcloud

When moving a card,the PUT method is not allowed.
Content of arg "description" triggers multiple rules.

### DIFF
--- a/plugins/nextcloud-rule-exclusions-before.conf
+++ b/plugins/nextcloud-rule-exclusions-before.conf
@@ -1338,6 +1338,21 @@ SecRule REQUEST_FILENAME "@rx /ocs/v[0-9]+\.php/apps/user_status/api/v[0-9]+/use
     setvar:'tx.allowed_methods=%{tx.allowed_methods} PUT DELETE'"
 
 #
+# [ Nextcloud Deck ]
+#
+
+# Moving a card in Deck app
+SecRule REQUEST_FILENAME "@rx /apps/deck/cards/[0-9]+/reorder$" \
+    "id:9508810,\
+    phase:1,\
+    pass,\
+    t:none,\
+    nolog,\
+    ver:'nextcloud-rule-exclusions-plugin/1.0.0',\
+    ctl:ruleRemoveTargetByTag=OWASP_CRS;ARGS:description,\
+    setvar:'tx.allowed_methods=%{tx.allowed_methods} PUT'"
+
+#
 # [ Users ]
 #
 
@@ -1915,18 +1930,3 @@ SecRule REQUEST_FILENAME "@rx /apps/mail/api/outbox/(?:from-draft/)?[0-9]+$" \
     ctl:ruleRemoveTargetByTag=OWASP_CRS;ARGS:json.body,\
     ctl:ruleRemoveTargetByTag=OWASP_CRS;ARGS:json.editorBody,\
     ctl:ruleRemoveTargetByTag=OWASP_CRS;ARGS:json.subject"
-
-#
-# [ Nextcloud Deck ]
-#
-
-# Moving a card in Deck app
-SecRule REQUEST_FILENAME "@rx /apps/deck/cards/[0-9]+/reorder$" \
-    "id:9508993,\
-    phase:1,\
-    pass,\
-    t:none,\
-    nolog,\
-    ver:'nextcloud-rule-exclusions-plugin/1.0.0',\
-    ctl:ruleRemoveTargetByTag=OWASP_CRS;ARGS:description,\
-    setvar:'tx.allowed_methods=%{tx.allowed_methods} PUT'"
--- a/plugins/nextcloud-rule-exclusions-before.conf
+++ b/plugins/nextcloud-rule-exclusions-before.conf
@@ -1916,8 +1916,12 @@ SecRule REQUEST_FILENAME "@rx /apps/mail/api/outbox/(?:from-draft/)?[0-9]+$" \
     ctl:ruleRemoveTargetByTag=OWASP_CRS;ARGS:json.editorBody,\
     ctl:ruleRemoveTargetByTag=OWASP_CRS;ARGS:json.subject"
 
-# Moving a card in Deck application
-SecRule REQUEST_FILENAME "@rx /apps/deck/cards/[0-9]+/reorder?$" \
+#
+# [ Nextcloud Deck ]
+#
+
+# Moving a card in Deck app
+SecRule REQUEST_FILENAME "@rx /apps/deck/cards/[0-9]+/reorder$" \
     "id:9508993,\
     phase:1,\
     pass,\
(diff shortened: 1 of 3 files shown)


tests/regression/nextcloud-rule-exclusions-plugin/9508993.yaml
```diff
@@ -0,0 +1,26 @@
...
+              Host: localhost
+              User-Agent: OWASP ModSecurity Core Rule Set
+              Accept: text/xml,application/xml,application/xhtml+xml,text/html;q=0.9,text/plain;q=0.8,image/png,*/*;q=0.5
+            port: 80
+            method: PUT
+            uri: /apps/deck/cards/1/reorder
+            data: "description=This is a description.\n\nUse this [link](https://github.com/microsoft/vscode/issues?page=3&q=is%3Aissue+is%3Aopen).\n"
+          output:
+            no_log_contains: |
+              id"911100"|id "93(?:[23]2[10]0)"|id "94(?:2[234][03467]0|9110)"|id "980130"
```
theseion, line 26: I appreciate the optimisation of the regular expression, however, in this situation readability is much more important. Please write use full rule numbers; using a single expression to match multiple rules with the last digit is fine IMO, e.g., `93223[156]`. There's also a missing space after the first `id`.

tests/regression/nextcloud-rule-exclusions-plugin/9508993.yaml
```diff
@@ -0,0 +1,26 @@
...
+          input:
+            dest_addr: 127.0.0.1
+            headers:
+              Host: localhost
+              User-Agent: OWASP ModSecurity Core Rule Set
+              Accept: text/xml,application/xml,application/xhtml+xml,text/html;q=0.9,text/plain;q=0.8,image/png,*/*;q=0.5
+            port: 80
+            method: PUT
+            uri: /apps/deck/cards/1/reorder
+            data: "description=This is a description.\n\nUse this [link](https://github.com/microsoft/vscode/issues?page=3&q=is%3Aissue+is%3Aopen).\n"
```
theseion, line 23: Use YAML scalar.
```suggestion
            data: |
              description=This is a description.
              
              
              Use this [link](https://github.com/microsoft/vscode/issues?page=3&q=is%3Aissue+is%3Aopen).
              

```

plugins/nextcloud-rule-exclusions-before.conf
```diff
@@ -1915,3 +1915,14 @@ SecRule REQUEST_FILENAME "@rx /apps/mail/api/outbox/(?:from-draft/)?[0-9]+$" \
     ctl:ruleRemoveTargetByTag=OWASP_CRS;ARGS:json.body,\
     ctl:ruleRemoveTargetByTag=OWASP_CRS;ARGS:json.editorBody,\
     ctl:ruleRemoveTargetByTag=OWASP_CRS;ARGS:json.subject"
+
+# Moving a card in Deck application
+SecRule REQUEST_FILENAME "@rx /apps/deck/cards/[0-9]+/reorder?$" \
```
theseion, line 1920: The question mark at the end makes only the `r` optional, which is probably not your intention.

theseion (changes requested):

theseion:
@EsadCetiner Would you mind taking a look as well?

plugins/nextcloud-rule-exclusions-before.conf
```diff
@@ -1915,3 +1915,14 @@ SecRule REQUEST_FILENAME "@rx /apps/mail/api/outbox/(?:from-draft/)?[0-9]+$" \
...
+    "id:9508993,\
```
EsadCetiner, line 1921: You've placed this rule underneath the Nextcloud Mail group, you should place Nextcloud Deck rule exclusions under it's own group. It's easier to understand what a group of rule does at a glance vs having to look into detail to each rule.

EsadCetiner (changes requested):

tests/regression/nextcloud-rule-exclusions-plugin/9508993.yaml
mivek: I listed all ids

tests/regression/nextcloud-rule-exclusions-plugin/9508993.yaml
mivek: done

plugins/nextcloud-rule-exclusions-before.conf
EsadCetiner: That's better, but you should use a different rule ID. There's only about 7 rule IDs left now. There's lots of free space if you look before the Users group, about 40 rule IDs free if you use ``9508810``. I doubt this will be the only false positive with Deck, knowing Nextcloud.

tests/regression/nextcloud-rule-exclusions-plugin/9508993.yaml
```diff
@@ -0,0 +1,29 @@
...
+            port: 80
+            method: PUT
+            uri: /apps/deck/cards/1/reorder
+            data: |
+              description=This is a description.
+
+              Use this [link](https://github.com/microsoft/vscode/issues?page=3&q=is%3Aissue+is%3Aopen).
+          output:
+            no_log_contains: |
+              id "911100"|id "932200"|id "933210"|id "942200"|id "942260"|id "942370"|id "942430"|id "942440"|id "949110"|id "980130"
```
EsadCetiner, line 29: @theseion 949110 and 980130 are for anomoly scoring and corelations respectively, is it worth it to test for these rules, at least for 949110 it should never trigger when the regression tests are running?

EsadCetiner (changes requested):

tests/regression/nextcloud-rule-exclusions-plugin/9508993.yaml
theseion: That's actually an interesting observation. I've been wandering recently about false positive tests. Negative matches are obviously a bit tricky, but we could just say: no rule should trigger. This is probably the case in 99% of the cases (at least for exclusion rule plugins). On the other hand, it could also mean that many test start failing at the same time, e.g., when a new rule is introduced that triggers on something.
Still, seeing the failure is probably more important. What do you think, should we discuss replacing individual rule matches with "no rules should trigger" for the Nextcloud plugin?

EsadCetiner: I've been wondering about that as well, Nextcloud isn't very CRS friendly and this will really make it easier to ensure there aren't any FPs with Nextcloud. It'll be much easier to maintain in the long run and cause fewer issues for end users. But I'm not sure how that would work, would you just write a regex rule that checks for any rule ID within the 911000-959999 range?
But for this test in particular, should we remove 949110 and 980130?

theseion: I think I'd do `no_log_contains: id "\d+"`. That would match __any__ rule. But yes, let's postpone that. I agree, let's remove those two matches to be consistent.

plugins/nextcloud-rule-exclusions-before.conf
mivek: I did change the rule ID

EsadCetiner (approved):

theseion (approved):

theseion:
Thank you @mivek! Thanks for the help @EsadCetiner.

mivek:
Glad to contribute, I am just starting to integrate crs with Nextcloud. I may open a couple more PR ^^

theseion:
Please do ;)

PR #60, opened 2024-02-19
chore: use flag in regexes for case insensitive matching

theseion:


RedXanadu (approved):
Makes sense in context. Simplified patterns are always better :slightly_smiling_face:

PR #61, opened 2024-03-02
fix: rules and tests after running at paranoia level 4

theseion:
With the CRS v4 images, it has become apparent that plugin tests were being run at paranoia level 1. Now, they are running at paranoia level 4, revealing some mistakes in the tests and rules.

### DIFF
--- a/plugins/nextcloud-rule-exclusions-before.conf
+++ b/plugins/nextcloud-rule-exclusions-before.conf
@@ -866,17 +866,14 @@ SecRule REQUEST_FILENAME "@endsWith /login/webauthn/finish" \
     t:none,\
     nolog,\
     ver:'nextcloud-rule-exclusions-plugin/1.0.0',\
-    ctl:ruleRemoveTargetById=920273;ARGS,\
-    ctl:ruleRemoveTargetById=932236;ARGS,\
-    ctl:ruleRemoveTargetById=942200;ARGS,\
-    ctl:ruleRemoveTargetById=942260;ARGS,\
-    ctl:ruleRemoveTargetById=942340;ARGS,\
-    ctl:ruleRemoveTargetById=942370;ARGS,\
-    ctl:ruleRemoveTargetById=942430;ARGS,\
-    ctl:ruleRemoveTargetById=942431;ARGS,\
-    ctl:ruleRemoveTargetById=942432;ARGS,\
+    ctl:ruleRemoveTargetById=920273;ARGS:data,\
+    ctl:ruleRemoveTargetById=932236;ARGS:data,\
+    ctl:ruleRemoveTargetById=920273;ARGS:json.data,\
+    ctl:ruleRemoveTargetById=932236;ARGS:json.data,\
     ctl:ruleRemoveTargetById=920272;REQUEST_BODY,\
-    ctl:ruleRemoveTargetById=920273;REQUEST_BODY"
+    ctl:ruleRemoveTargetById=920273;REQUEST_BODY,\
+    ctl:ruleRemoveTargetByTag=attack-sqli;ARGS:data,\
+    ctl:ruleRemoveTargetByTag=attack-sqli;ARGS:json.data"
 
 # Logging in via desktop app
 SecRule REQUEST_FILENAME "@rx /login/v[0-9\.]+/grant$" \
@@ -1462,17 +1459,13 @@ SecRule REQUEST_FILENAME "@endsWith /settings/api/personal/webauthn/registration
     t:none,\
     nolog,\
     ver:'nextcloud-rule-exclusions-plugin/1.0.0',\
-    ctl:ruleRemoveTargetById=920273;ARGS,\
-    ctl:ruleRemoveTargetById=932236;ARGS,\
-    ctl:ruleRemoveTargetById=942200;ARGS,\
-    ctl:ruleRemoveTargetById=942260;ARGS,\
-    ctl:ruleRemoveTargetById=942340;ARGS,\
-    ctl:ruleRemoveTargetById=942370;ARGS,\
-    ctl:ruleRemoveTargetById=942421;ARGS,\
-    ctl:ruleRemoveTargetById=942430;ARGS,\
-    ctl:ruleRemoveTargetById=942431;ARGS,\
-    ctl:ruleRemoveTargetById=942432;ARGS,\
-    ctl:ruleRemoveTargetById=920273;REQUEST_BODY"
+    ctl:ruleRemoveTargetById=920273;ARGS:data,\
+    ctl:ruleRemoveTargetById=932236;ARGS:data,\
+    ctl:ruleRemoveTargetById=920273;ARGS:json.data,\
+    ctl:ruleRemoveTargetById=932236;ARGS:json.data,\
+    ctl:ruleRemoveTargetById=920273;REQUEST_BODY,\
+    ctl:ruleRemoveTargetByTag=attack-sqli;ARGS:data,\
+    ctl:ruleRemoveTargetByTag=attack-sqli;ARGS:json.data"
 
 # When a user is changing their password under settings --> Personal --> Security --> Password
 SecRule REQUEST_FILENAME "@endsWith /settings/personal/changepassword" \
--- a/plugins/nextcloud-rule-exclusions-before.conf
+++ b/plugins/nextcloud-rule-exclusions-before.conf
@@ -263,6 +263,7 @@ SecRule REQUEST_FILENAME "@rx /remote\.php/dav/comments/files(?:/[0-9]+)+$" \
     t:none,\
     nolog,\
     ver:'nextcloud-rule-exclusions-plugin/1.0.0',\
+    ctl:ruleRemoveTargetByTag=OWASP_CRS;ARGS:message,\
     ctl:ruleRemoveTargetByTag=OWASP_CRS;ARGS:json.message,\
     setvar:'tx.allowed_request_content_type=%{tx.allowed_request_content_type} |text/plain|'"
 
@@ -287,8 +288,8 @@ SecRule REQUEST_URI "@rx /apps/text/(?:public/)?session/sync$" \
     ctl:ruleRemoveTargetById=920272;REQUEST_BODY,\
     ctl:ruleRemoveTargetById=920273;REQUEST_BODY,\
     ctl:ruleRemoveTargetById=921110;REQUEST_BODY,\
-    ctl:ruleRemoveTargetByTag=OWASP_CRS;ARGS:json.autosaveContent,\
-    ctl:ruleRemoveTargetByTag=OWASP_CRS;ARGS:autosaveContent"
+    ctl:ruleRemoveTargetByTag=OWASP_CRS;ARGS:autosaveContent,\
+    ctl:ruleRemoveTargetByTag=OWASP_CRS;ARGS:json.autosaveContent"
 
 # Text app content can be anything
 # We disable CRS for all ARGS since the paramater keeps on changing and can't be predicted
@@ -337,6 +338,7 @@ SecRule REQUEST_FILENAME "@rx /ocs/v[0-9]\.php/apps/files_sharing/api/v[0-9]/sha
     pass,\
     t:none,\
     nolog,\
+    ctl:ruleRemoveTargetByTag=OWASP_CRS;ARGS:password,\
     ctl:ruleRemoveTargetByTag=OWASP_CRS;ARGS:json.password,\
     ver:'nextcloud-rule-exclusions-plugin/1.0.0'"
 
@@ -554,8 +556,12 @@ SecRule REQUEST_FILENAME "@rx /apps/text/(?:public/)?(?:session/(?:sync|close|pu
     nolog,\
     ver:'nextcloud-rule-exclusions-plugin/1.0.0',\
     chain"
-    SecRule ARGS:json.sessionToken "@rx ^(?i)[a-z0-9+/]+$" \
+    SecRule ARGS:sessionToken|ARGS:json.sessionToken "@rx ^(?i)[a-z0-9+/]+$" \
         "t:none,\
+        ctl:ruleRemoveTargetById=920273;ARGS:sessionToken,\
+        ctl:ruleRemoveTargetById=932236;ARGS:sessionToken,\
+        ctl:ruleRemoveTargetById=942432;ARGS:sessionToken,\
+        ctl:ruleRemoveTargetById=942450;ARGS:sessionToken,\
         ctl:ruleRemoveTargetById=920273;ARGS:json.sessionToken,\
         ctl:ruleRemoveTargetById=932236;ARGS:json.sessionToken,\
         ctl:ruleRemoveTargetById=942432;ARGS:json.sessionToken,\
@@ -570,6 +576,13 @@ SecRule REQUEST_FILENAME "@rx /apps/text/(?:public/)?session/(?:sync|close|push)
     t:none,\
     nolog,\
     ver:'nextcloud-rule-exclusions-plugin/1.0.0',\
+    ctl:ruleRemoveTargetById=932236;ARGS:documentState,\
+    ctl:ruleRemoveTargetById=932250;ARGS:documentState,\
+    ctl:ruleRemoveTargetById=920273;ARGS:documentState,\
+    ctl:ruleRemoveTargetById=941100;ARGS:documentState,\
+    ctl:ruleRemoveTargetById=942210;ARGS:documentState,\
+    ctl:ruleRemoveTargetById=942432;ARGS:documentState,\
+    ctl:ruleRemoveTargetById=942450;ARGS:documentState,\
     ctl:ruleRemoveTargetById=932236;ARGS:json.documentState,\
     ctl:ruleRemoveTargetById=932250;ARGS:json.documentState,\
     ctl:ruleRemoveTargetById=920273;ARGS:json.documentState,\
@@ -608,12 +621,17 @@ SecRule REQUEST_FILENAME "@rx /apps/text/(?:public/)?session/push$" \
     nolog,\
     ver:'nextcloud-rule-exclusions-plugin/1.0.0',\
     chain"
-    SecRule ARGS:awareness "@rx ^(?i)[a-z0-9=+/]+$" \
+    SecRule ARGS:awareness|ARGS:json.awareness "@rx ^(?i)[a-z0-9=+/]+$" \
         "t:none,\
+        ctl:ruleRemoveTargetById=920273;ARGS:awareness,\
+        ctl:ruleRemoveTargetById=932236;ARGS:awareness,\
+        ctl:ruleRemoveTargetById=942432;ARGS:awareness,\
+        ctl:ruleRemoveTargetById=942450;ARGS:awareness,\
         ctl:ruleRemoveTargetById=920273;ARGS:json.awareness,\
         ctl:ruleRemoveTargetById=932236;ARGS:json.awareness,\
         ctl:ruleRemoveTargetById=942432;ARGS:json.awareness,\
-        ctl:ruleRemoveTargetById=942450;ARGS:json.awareness"
+        ctl:ruleRemoveTargetById=942450;ARGS:json.awareness,\
+        ctl:ruleRemoveTargetById=920273;REQUEST_BODY"
 
 # Checking for attachemnts on public shares
 SecRule REQUEST_FILENAME "@endsWith /apps/text/attachments" \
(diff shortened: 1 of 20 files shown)


tests/regression/nextcloud-rule-exclusions-plugin/9508311.yaml
```diff
@@ -30,6 +31,7 @@ tests:
               Host: localhost
               User-Agent: OWASP CRS
               Accept: text/xml,application/xml,application/xhtml+xml,text/html;q=0.9,text/plain;q=0.8,image/png,*/*;q=0.5
+              Content-Type: application/json
```
EsadCetiner, line 34: All of them are json?

theseion: Should be fixed now.

theseion:
@EsadCetiner ready for review.

tests/regression/nextcloud-rule-exclusions-plugin/9508991.yaml
```diff
@@ -39,10 +41,12 @@ tests:
               Host: localhost
               User-Agent: OWASP CRS
               Accept: text/xml,application/xml,application/xhtml+xml,text/html;q=0.9,text/plain;q=0.8,image/png,*/*;q=0.5
+              Content-Type: application/json
             port: 80
             method: PUT
             uri: /apps/mail/api/drafts/1
-            data: "json.data.body.value=<script>"
+            data: |
+              {"data.body.value": "<script>"}
```
azurit, line 49: This should probably rewrites into `{"data": {"body": {"value": "<script>"}}}`

tests/regression/nextcloud-rule-exclusions-plugin/9508991.yaml
```diff
@@ -119,10 +129,12 @@ tests:
               Host: localhost
               User-Agent: OWASP CRS
               Accept: text/xml,application/xml,application/xhtml+xml,text/html;q=0.9,text/plain;q=0.8,image/png,*/*;q=0.5
+              Content-Type: application/json
             port: 80
             method: PUT
             uri: /apps/mail/api/drafts/{id}
-            data: "json.data.body.value=<script>"
+            data: |
+              {"data.body.value": "<script>"}
```
azurit, line 137: This should probably rewrites into `{"data": {"body": {"value": "<script>"}}}`

tests/regression/nextcloud-rule-exclusions-plugin/9508991.yaml
```diff
@@ -139,10 +151,12 @@ tests:
               Host: localhost
               User-Agent: OWASP CRS
               Accept: text/xml,application/xml,application/xhtml+xml,text/html;q=0.9,text/plain;q=0.8,image/png,*/*;q=0.5
+              Content-Type: application/json
             port: 80
             method: PUT
             uri: /apps/mail/api/drafts/{id}
-            data: "json.data.editorBody=<script>"
+            data: |
+              {"data.editorBody": "<script>"}
```
azurit, line 159: This should probably rewrites into `{"data": {"editorBody": "<script>"}}`

tests/regression/nextcloud-rule-exclusions-plugin/9508991.yaml
```diff
@@ -59,10 +63,12 @@ tests:
               Host: localhost
               User-Agent: OWASP CRS
               Accept: text/xml,application/xml,application/xhtml+xml,text/html;q=0.9,text/plain;q=0.8,image/png,*/*;q=0.5
+              Content-Type: application/json
             port: 80
             method: PUT
             uri: /apps/mail/api/drafts/1
-            data: "json.data.editorBody=<script>"
+            data: |
+              {"data.editorBody": "<script>"}
```
azurit, line 71: This should probably rewrites into {"data": {"editorBody": "<script>"}}

tests/regression/nextcloud-rule-exclusions-plugin/9508400.yaml
```diff
@@ -52,7 +52,7 @@ tests:
             uri: /login
             data: "redirect_url=/login/v2/grant?user=Esad%20Cetiner&stateToken=sample-token"
           output:
-            no_log_contains: |
+            no_log_contains: |-
               id "920230"|id "932(?:19|20)0"|id "94243[12]"
```
azurit, line 56: Why we need doing this in tests? It's unreadable, cannot be searched for rule IDs and so on. Let's simply specify whole rule IDs without using regexes.

azurit:
@theseion Why are you adding json and also non-json variants of arguments everywhere?

theseion:
In v2, JSON variables _are not_ prefixed with `json`. So, for v2: `ARGS:sessionToken`, for v3: `ARGS:json:sessionToken`.

plugins/nextcloud-rule-exclusions-before.conf
```diff
@@ -834,15 +866,17 @@ SecRule REQUEST_FILENAME "@endsWith /login/webauthn/finish" \
...
-    ctl:ruleRemoveTargetById=942432;ARGS:json.data"
+    ctl:ruleRemoveTargetById=920273;ARGS,\
+    ctl:ruleRemoveTargetById=932236;ARGS,\
+    ctl:ruleRemoveTargetById=942200;ARGS,\
+    ctl:ruleRemoveTargetById=942260;ARGS,\
+    ctl:ruleRemoveTargetById=942340;ARGS,\
+    ctl:ruleRemoveTargetById=942370;ARGS,\
+    ctl:ruleRemoveTargetById=942430;ARGS,\
+    ctl:ruleRemoveTargetById=942431;ARGS,\
+    ctl:ruleRemoveTargetById=942432;ARGS,\
```
EsadCetiner, line 877: Why are you disabling these rules for all ARGS?

plugins/nextcloud-rule-exclusions-before.conf
```diff
@@ -1422,16 +1462,17 @@ SecRule REQUEST_FILENAME "@endsWith /settings/api/personal/webauthn/registration
...
+    ctl:ruleRemoveTargetById=932236;ARGS,\
+    ctl:ruleRemoveTargetById=942200;ARGS,\
+    ctl:ruleRemoveTargetById=942260;ARGS,\
+    ctl:ruleRemoveTargetById=942340;ARGS,\
+    ctl:ruleRemoveTargetById=942370;ARGS,\
+    ctl:ruleRemoveTargetById=942421;ARGS,\
+    ctl:ruleRemoveTargetById=942430;ARGS,\
+    ctl:ruleRemoveTargetById=942431;ARGS,\
+    ctl:ruleRemoveTargetById=942432;ARGS,\
+    ctl:ruleRemoveTargetById=920273;REQUEST_BODY"
```
EsadCetiner, line 1475: Why are you disabling these rules for all ARGS?

plugins/nextcloud-rule-exclusions-before.conf
theseion: Because, unfortunately, keys aren't hierarchical. For example, for the structure `{"one": "{"two": "three"}}`, ModSecurity will create the arg `one.two` with value `three` but no key `one` will exist. It's also not possible to match with a regular expression (at least it didn't work when I tried).

For a case like this, there are two options:
- exclude all `ARGS`
- list all possible keys

I wasn't sure that the object in the test contained all possible keys, so I opted for excluding `ARGS`.

EsadCetiner: The data looks like this, I've just redacted the actual data while preserving the general structure.
```
{"data":"{\"id\":\"ZZ7gbdsfkK-BUODFAbiujsdf7-is\",\"type\":\"public-key\",\"rawId\":\"h7GEJH79Ds+ub86g8UGAS+Hs=\",\"response\":{\"authenticatorData\":\"3/zX+yiugsdf678567GYUSHADF/GQ+r657HKJGSFAD756==\",\"clientDataJSON\":\"iyuhbsbbnl68567856876tghyiYUVUVASFDYUV79bnugiyuhbsbbnl68567856876tghyiYUVUVASFDYUV79bnugiyuhbsbbnl68567856876tghyiYUVUVASFDYUV79bnugiyuhbsbbnl68567856876tghyiYUVUVASFDYUV79bnugiyuhbsbbnl68567856876tghyiYUVUVASFDYUV79bnug=\",\"signature\":\"yb8g869gBHJKAFSHJ+yYUUV7855687+Kq+ldfs86\",\"userHandle\":\"UHGASF=\"}}"}
```
So if I'm understanding this correctly, if there was a false positive within rawId the matched var would be ``json.data.rawId``? This doesn't line up with what's shown in my logs, the matched variable is ``json.data`` and not something else.
Maybe it's a difference between V2 and V3?

azurit: > if there was a false positive within rawId the matched var would be `json.data.rawId`?

Yes. Can you show your logs?

EsadCetiner:
```
2024/03/03 23:34:50 [info] 3826458#3826458: *53 ModSecurity: Warning. Matched "Operator `Rx' with parameter `((?:[~!@#\$%\^&\*\(\)\-\+=\{\}\[\]\|:;\"'\xc2\xb4\xe2\x80\x99\xe2\x80\x98`<>][^~!@#\$%\^&\*\(\)\-\+=\{\}\[\]\|:;\"'\xc2\xb4\xe2\x80\x99\xe2\x80\x98`<>]*?){12})' against variable `ARGS:json.data' (Value: `{"id":"ZZ7gbdsfkK-BUODFAbiujsdf7-is","type":"public-key","rawId":"h7GEJH79Ds+ub86g8UGAS (458 characters omitted)' ) [file "/etc/nginx/modsecurity/coreruleset/rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf"] [line "1296"] [id "942430"] [rev ""] [msg "Restricted SQL Character Anomaly Detection (args): # of special characters exceeded (12)"] [data "Matched Data: {\x22id\x22:\x22ZZ7gbdsfkK-BUODFAbiujsdf7-is\x22,\x22type\x22:\x22 found within ARGS:json.data: {\x22id\x22:\x2ZZ7gbdsfkK-BUODFAbiujsdf7-is\x22,\x22type\x22:\x22public-key\x22,\x22rawId\x22:\x22Zc6BjnF9TU dH7H (462 characters omitted)"] [severity "4"] [ver "OWASP_CRS/4.0.0"] [maturity "0"] [accuracy "0"] [tag "application-multi"] [tag "language-multi"] [tag "platform-multi"] [tag "attack-sqli"] [tag "OWASP_CRS"] [tag "capec/1000/152/248/66"] [tag "PCI/6.5.2"] [tag "paranoia-level/2"] [hostname "0.0.0.0"] [uri "/login/webauthn/finish"] [unique_id "170946929077.753528"] [ref "o0,60o0,60v10,558t:urlDecodeUni"], client: 0.0.0.0, server: cloud.example.com, request: "POST /login/webauthn/finish HTTP/2.0", host: "cloud.example.com"
```

azurit: Can you show full audit log for that request?

EsadCetiner: I'm not sure what your looking for, libModSecurity3 doesn't log the request body. Everything else is just different rules that are triggered.
```
---7JWi0PR3---A--
[03/Mar/2024:23:34:50 +1100] 170946929077.753528 0.0.0.0 54272 0.0.0.0 443
---7JWi0PR3---B--
POST /login/webauthn/finish HTTP/2.0
sec-gpc: 1
user-agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/122.0.0.0 Safari/537.36
sec-fetch-site: same-origin
sec-ch-ua-mobile: ?0
origin: https://cloud.example.com
content-type: application/json
requesttoken: 42hg+fdshgfdh=:54/dfgh=
sec-ch-ua-platform: "Windows"
accept: application/json, text/plain, */*
sec-ch-ua: "Chromium";v="122", "Not(A:Brand";v="24", "Brave";v="122"
accept-language: en-US,en;q=0.8
content-length: 599
host: cloud.example.com
sec-fetch-mode: cors
sec-fetch-dest: empty
accept-encoding: gzip, deflate, br
```

Do you want all of the triggered rules as well?

azurit: > libModSecurity3 doesn't log the request body

Are you sure? This is not possible to debug without request body.

EsadCetiner: Yes I'm sure, it does log the response body though.
But isn't what I gave earlier enough? that's the payload sent by the client (according to chrome's inspect element)
Otherwise, I could try finding the full request with wireshark.

azurit: I just noticed that the data is really strange - it looks like that Nextcloud is sending JSON data as a string inside a JSON:
`{"data":"{\"id\":\"ZZ7gbdsfkK-BUODFAbiujsdf7-is\",\"type\":\"public-key\",\"rawId\":\"h7GEJH79Ds+ub86g8UGAS+Hs=\",\"response\":{\"authenticatorData\":\"3/zX+yiugsdf678567GYUSHADF/GQ+r657HKJGSFAD756==\",\"clientDataJSON\":\"iyuhbsbbnl68567856876tghyiYUVUVASFDYUV79bnugiyuhbsbbnl68567856876tghyiYUVUVASFDYUV79bnugiyuhbsbbnl68567856876tghyiYUVUVASFDYUV79bnugiyuhbsbbnl68567856876tghyiYUVUVASFDYUV79bnugiyuhbsbbnl68567856876tghyiYUVUVASFDYUV79bnug=\",\"signature\":\"yb8g869gBHJKAFSHJ+yYUUV7855687+Kq+ldfs86\",\"userHandle\":\"UHGASF=\"}}"}`

Notice that everything inside value for "data" element is between `"` and all other `"` insde the data are escaped.

theseion: Yes, that's what I'm seeing. Because the tests now use proper json structure, the output looks different from yours. The only explanation is what @azurit suggests.

theseion: I'll update the tests and the rules.

azurit: @theseion Cool! I also suggest replacing all 942* exclusions (there are too many, it's not worth it) with this:
`ctl:ruleRemoveTargetByTag=attack-sqli;ARGS:json.data`

EsadCetiner (approved):

EsadCetiner:
@theseion Thank you for cleaning all of this up!

plugins/nextcloud-rule-exclusions-before.conf
```diff
@@ -554,12 +556,17 @@ SecRule REQUEST_FILENAME "@rx /apps/text/(?:public/)?(?:session/(?:sync|close|pu
     nolog,\
     ver:'nextcloud-rule-exclusions-plugin/1.0.0',\
     chain"
-    SecRule ARGS:json.sessionToken "@rx ^(?i)[a-z0-9+/]+$" \
+    SecRule ARGS:sessionToken|ARGS:json.sessionToken "@rx ^(?i)[a-z0-9+/]+$" \
```
azurit, line 559: As this regex is technically correct i would move `(?i)` before `^` as it's probably meant to apply for the whole regex..

plugins/nextcloud-rule-exclusions-before.conf
```diff
@@ -554,12 +556,17 @@ SecRule REQUEST_FILENAME "@rx /apps/text/(?:public/)?(?:session/(?:sync|close|pu
...
+        ctl:ruleRemoveTargetById=920273;ARGS:sessionToken,\
+        ctl:ruleRemoveTargetById=932236;ARGS:sessionToken,\
+        ctl:ruleRemoveTargetById=942432;ARGS:sessionToken,\
+        ctl:ruleRemoveTargetById=942450;ARGS:sessionToken,\
         ctl:ruleRemoveTargetById=920273;ARGS:json.sessionToken,\
         ctl:ruleRemoveTargetById=932236;ARGS:json.sessionToken,\
         ctl:ruleRemoveTargetById=942432;ARGS:json.sessionToken,\
-        ctl:ruleRemoveTargetById=942450;ARGS:json.sessionToken"
+        ctl:ruleRemoveTargetById=942450;ARGS:json.sessionToken,\
+        ctl:ruleRemoveTargetById=920273;REQUEST_BODY"
```
azurit, line 569: Do we really want to exclude this only if `sessionToken` matches regex above?

EsadCetiner: The idea was to use character whitelisting to make the rule exclusion stricter.

azurit: I understand that part but whitelisting `REQUEST_BODY` only if some header matches some regex doesn't make sense to me. Why does `REQUEST_BODY` matches rule `920273` only if `sessionToken` matches that regex? How is that connected?

theseion: 920273 checks `ARGS` and `REQUEST_BODY`. So we only need to exclude the request body in this particular case.

azurit: > 920273 checks `ARGS` and `REQUEST_BODY`. So we only need to exclude the request body in this particular case.

Unfortunately it doesn't work like that. We (may, it depends) need both for correct exclusion.

azurit (approved):

plugins/nextcloud-rule-exclusions-before.conf
azurit: I'm marking as resolved to not block merging this PR.

plugins/nextcloud-rule-exclusions-before.conf
```diff
@@ -834,15 +866,14 @@ SecRule REQUEST_FILENAME "@endsWith /login/webauthn/finish" \
     t:none,\
     nolog,\
     ver:'nextcloud-rule-exclusions-plugin/1.0.0',\
+    ctl:ruleRemoveTargetById=920273;ARGS:data,\
+    ctl:ruleRemoveTargetById=932236;ARGS:data
```
EsadCetiner, line 870:
```suggestion
    ctl:ruleRemoveTargetById=932236;ARGS:data,\
```

EsadCetiner (changes requested):

plugins/nextcloud-rule-exclusions-before.conf
```diff
@@ -834,15 +866,14 @@ SecRule REQUEST_FILENAME "@endsWith /login/webauthn/finish" \
...
     ctl:ruleRemoveTargetById=920273;ARGS:json.data,\
-    ctl:ruleRemoveTargetById=932236;ARGS:json.data,\
-    ctl:ruleRemoveTargetById=942200;ARGS:json.data,\
-    ctl:ruleRemoveTargetById=942260;ARGS:json.data,\
-    ctl:ruleRemoveTargetById=942340;ARGS:json.data,\
-    ctl:ruleRemoveTargetById=942370;ARGS:json.data,\
-    ctl:ruleRemoveTargetById=942430;ARGS:json.data,\
-    ctl:ruleRemoveTargetById=942431;ARGS:json.data,\
-    ctl:ruleRemoveTargetById=942432;ARGS:json.data"
+    ctl:ruleRemoveTargetById=932236;ARGS:json.data
```
EsadCetiner: another one
```suggestion
    ctl:ruleRemoveTargetById=932236;ARGS:json.data,\
```

EsadCetiner (changes requested):

EsadCetiner (approved):

plugins/nextcloud-rule-exclusions-before.conf
theseion: You mean because there may be other cases where we need to exclude the body? Sorry, I can't follow your logic.

theseion: I want to cut release 1.1.0 of the plugin, so if you want to change something before that, let me know.

azurit: > You mean because there may be other cases where we need to exclude the body? Sorry, I can't follow your logic.

Because we are excluding for specific arguments / variables. If you exclude a rule for `REQUEST_BODY` it won't stop mod_security to apply that rule to other variables.

azurit: > I want to cut release 1.1.0 of the plugin, so if you want to change something before that, let me know.

No, go ahead!

theseion: > Because we are excluding for specific arguments / variables. If you exclude a rule for REQUEST_BODY it won't stop mod_security to apply that rule to other variables.

I understand that. Well, let's discuss it when it comes up again :)

PR #62, opened 2024-03-07
feat: add support for Nextcloud 28

EsadCetiner:
Nextcloud 28 has made a fair few changes and some of these have resulted in new false positives. Most of these false positives are at PL-2 or higher or they they affect features that will only be used on occasion.


plugins/nextcloud-rule-exclusions-before.conf
```diff
@@ -1575,6 +1581,53 @@ SecRule REQUEST_FILENAME "@endsWith /index.php/apps/richdocuments/ajax/admin.php
...
+    "id:9508637,\
+    phase:1,\
+    pass,\
+    t:none,\
+    nolog,\
+    ver:'nextcloud-rule-exclusions-plugin/1.1.0',\
+    ctl:ruleRemoveTargetById=932236;XML:/*,\
+    ctl:ruleRemoveTargetById=942430;XML:/*,\
+    ctl:ruleRemoveTargetById=942431;XML:/*,\
+    ctl:ruleRemoveTargetById=942432;XML:/*"
```
azurit, line 1629: Should be sufficient to use only `XML` instead of `XML:/*`.

azurit:
@EsadCetiner What is the purpose of splitting exclusions into rules 9508311, 9508312, 9508313 and 9508314? Seems like all of them are matching the same URLs.

EsadCetiner:
@azurit Those rule exclude rules for individual token used in Nextcloud's markdown text editor. All of these rule exclusions could be done in one rule, but then you won't be able to easily implement character whitelisting (which is currently the case).

plugins/nextcloud-rule-exclusions-before.conf
EsadCetiner: yup that works fine, is this a performance change?

azurit: Probably yes as it will perform an xpath search in XML data.

EsadCetiner: Looks like the linter didn't like that, should I revert the change?

EsadCetiner:
@azurit everything good to merge?

EsadCetiner:
@theseion I'm trying to use the ``REQUEST_BODY_LENGTH`` variable to check if there's no response body then disable rule 200002 since that's causing the XML parser to fail, but the linter doesn't think that's a valid variable. It's supported on both ModSec v2/v3 and Coraza so I don't know why it's marked as invalid.

theseion:
Wow, wouldn't have expected that. But your right, there's a mistake in the parser model.

I've opened an issue: https://github.com/coreruleset/secrules_parsing/issues/70.
@airween, any idea how to fix that one? I wasn't able to.

airween:
There is still a workflow [issue](https://github.com/coreruleset/nextcloud-rule-exclusions-plugin/actions/runs/8529308350/job/23364805222?pr=62#step:5:12), but I'm afraid that's an `ftw` problem. @theseion could you take a look at it?

plugins/nextcloud-rule-exclusions-before.conf
```diff
@@ -792,6 +843,63 @@ SecRule REQUEST_FILENAME "@rx /remote\.php/dav/calendars/[^/]+/trashbin/objects/
...
+    phase:1,\
+    pass,\
+    t:none,\
+    nolog,\
+    ver:'nextcloud-rule-exclusions-plugin/1.1.0',\
+    chain"
+    SecRule REQUEST_METHOD "@pm DELETE" \
+        "t:none,\
+        chain"
+        SecRule &REQUEST_BODY_LENGTH "@eq 1" \
```
airween, line 881: I'm not sure this is a valid expression, I mean `&REQUEST_BODY_LENGTH` makes no sense. `REQUEST_BODY_LENGTH` provides the length of request body - but what could mean `&` as a prefix?

And looks like Apache does not allow this, see the CI [error](https://github.com/coreruleset/nextcloud-rule-exclusions-plugin/actions/runs/8529308350/job/23364805222?pr=62#step:5:12).

airween (changes requested):

plugins/nextcloud-rule-exclusions-before.conf
EsadCetiner: I was thinking it meant 0 bytes in the request body, but I now realize that could also mean the number of request bodies in a request(afaik you can't have more than 1 request body, I'm not sure how that would even work). An example isn't provided [in the docs](https://github.com/owasp-modsecurity/ModSecurity/wiki/Reference-Manual-(v3.x)#user-content-REQUEST_BODY_LENGTH) so that's probably part of my confusion. I think the correct value here is ``REQUEST_BODY_LENGTH "@eq 0"`` but I'll be testing to make sure.

airween: As I see libmodsecurity3 allows this syntax, and the value (of `&REQUEST_BODY_LENGTH`) will be evaluated as `1` (even if the value of `REQUEST_BODY_LENGTH` is for eg. `12`).

I suggest you use `&REQUEST_BODY` or `REQUEST_BODY_LENGTH`, but not `&REQUEST_BODY_LENGTH`.

Note: it's not relevant here but the Apache's error message is a bit misleading: `Error creating rule: The & modificator does not apply to non-collection variables.`, because `REQUEST_BODY` is not a collection, but it allows to use `&` with that.

EsadCetiner: ``REQUEST_BODY_LENGTH "@eq 0"`` does it, everything should be fixed now. Thanks for clearing up my confusion.

theseion:
Yes, I know. I was waiting for your last PR to the parser to be merged and a new version to be published.

theseion:
All the checks are good now. @airween, you need to approve the changes, then we can merge.

airween (approved):
LGTM

PR #67, opened 2024-03-22
fix: requesting a static file at pl-2

EsadCetiner:
Fixes false positives when requesting a static file at pl-2 or higher with the ver argument. I haven't done any character whitelisting since this is just used for static files, there's nothing to attack.

### DIFF
--- a/plugins/nextcloud-rule-exclusions-before.conf
+++ b/plugins/nextcloud-rule-exclusions-before.conf
@@ -998,7 +998,12 @@ SecRule REQUEST_FILENAME "@rx /ocs/v[0-9]\.php/cloud/users(?:/[^/]+)?$" \
     ctl:ruleRemoveTargetByTag=OWASP_CRS;ARGS:json.password,\
     ctl:ruleRemoveTargetByTag=OWASP_CRS;ARGS:json.value"
 
+#
+# [ General rule exclusions ]
+#
+
 # Fix Nextcloud Cookie FPs
+# Fix false positives with static files version
 # Operator @unconditionalMatch is used instead of a SecAction because of a bug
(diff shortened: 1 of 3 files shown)


azurit:
@EsadCetiner How are you enforing that this exclusion is used only for static files?

EsadCetiner:
@azurit I'm not, the variable is only used for static files. Do you want me to write another rule that makes that check?

azurit:
@EsadCetiner I would better do it. I suggest to restrict it based on file extension.

EsadCetiner:
@azurit everything good to merge now?

plugins/nextcloud-rule-exclusions-before.conf
```diff
@@ -1124,6 +1128,16 @@ SecRule REQUEST_FILENAME "@unconditionalMatch" \
     ctl:ruleRemoveTargetById=942210;REQUEST_COOKIES:oc_sessionPassphrase,\
     ctl:ruleRemoveTargetById=942450;REQUEST_COOKIES:oc_sessionPassphrase"
 
+# Fix false positives with static files
+SecRule REQUEST_FILENAME "@rx \.(?:avif|bmp|brotli|bvr|css|eot|gif|gz|ico|jpeg|jpg|map|m?js|mp[34]|otf|png|svg|swf|ts|ttf|txt|wav|webp|woff2?)$" \
+    "id:9508511,\
+    phase:1,\
+    pass,\
+    t:none,\
```
azurit, line 1136: I suggest to add `t:lowercase`.

azurit: Or maybe better, prefix regex with `(?i)`.

azurit (approved):

PR #69, opened 2024-03-23
fix: .well-known/webdav fp

YeapGuy:
Adds a rule similar to 9508602 & 9508603 to prevent blocking discovering WebDAV on /.well-known:

```
ModSecurity: Warning. Matched "Operator `Within' with parameter `GET HEAD POST OPTIONS PUT' against variable `REQUEST_METHOD' (Value: `PROPFIND' ) [file "/usr/local/nginx/conf/conf.d/include/coreruleset/rules/REQUEST-911-METHOD-ENFORCEMENT.conf"] [line "28"] [id "911100"] [rev ""] [msg "Method is not allowed by policy"] [data "PROPFIND"] [severity "2"] [ver "OWASP_CRS/4.0.0"] [maturity "0"] [accuracy "0"] [tag "application-multi"] [tag "language-multi"] [tag "platform-multi"] [tag "attack-generic"] [tag "paranoia-level/1"] [tag "OWASP_CRS"] [tag "capec/1000/210/272/220/274"] [tag "PCI/12.1"] [hostname "172.19.0.7"] [uri "/.well-known/webdav/xxx/"] [unique_id "171119327475.703982"] [ref "v0,8"]
```

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -11,6 +11,7 @@
 - [ignatiev](https://github.com/ignatiev)
 - [Joost de Keijzer](https://github.com/joostdekeijzer)
 - [Max Leske](https://github.com/theseion)
+- [Matus](https://github.com/yeapguy)
 - [pyllyukko](https://github.com/pyllyukko)
 - [Chaim Sanders](https://github.com/csanders-git)
 - [Federico G. Schwindt](https://github.com/fgsch)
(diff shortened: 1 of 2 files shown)


EsadCetiner:
@YeapGuy Can you add your name to [the contributors file](https://github.com/coreruleset/nextcloud-rule-exclusions-plugin/blob/main/CONTRIBUTORS.md), it's sorted alphabetically by last name. And if you can write a test for the new rule then that would be great!

YeapGuy:
Rules 9508602 & 9508603 (which are very similar) don't have any tests, so I thought this rule doesn't need one either. (Honestly, I planned on copying and modifying the existing tests, because I don't feel capable to write a good test on my own 😅). Should I try to write one anyway?

EsadCetiner:
@YeapGuy Tests aren't a hard requirement for plugins at the moment, but it's good to write tests so we can catch regressions if we have to make changes for whatever reason. If you feel like you can't write a test then that's fine, but I'm happy to help if you want to. The new rule you introduced is pretty simple so it shouldn't be a problem.

EsadCetiner (approved):

EsadCetiner:
@YeapGuy Thank you for your contribution

PR #72, opened 2024-06-04
fix: fp regression with xml content

EsadCetiner:
A previous pull request removed XPath XML, this has caused a regression with XML content and introduced some new false positives. This PR reverts those changes


EsadCetiner:
@theseion It's already broken, that specific change was a suggestion to improve performance.

theseion (approved):

PR #79, opened 2024-06-15
fix: sqli fp with tokens and special characters

EsadCetiner:


tests/regression/nextcloud-rule-exclusions-plugin/9508312.yaml
```diff
@@ -537,3 +537,22 @@ tests:
           output:
             no_log_contains: |-
               id "920273"|id "942432"|id "942450"
+  - test_title: 9508312-28
```
theseion, line 540:
```suggestion
  - test_title: 9508312-29
```

tests/regression/nextcloud-rule-exclusions-plugin/9508420.yaml
theseion, line 1: Don't you want to keep these tests? I'd just move them to the new rule.

theseion (changes requested):

tests/regression/nextcloud-rule-exclusions-plugin/9508420.yaml
EsadCetiner: Most of these tests are testing the same payload, just against different URL paths. I did notice I missed one payload so I'll add that.

theseion (approved):

PR #82, opened 2024-07-11
fix: use case insensitivity for public shares when entering passwords

EsadCetiner:
Nextcloud has recently renamed the endpoint for entering passwords in public shares from ``/s/random/authenticate/showShare`` to ``/s/random/authenticate/showshare`` causing the rule that handles false positives with passwords to no longer match. This PR makes that rule case insensitive.

### DIFF
--- a/tests/regression/nextcloud-rule-exclusions-plugin/9508171.yaml
+++ b/tests/regression/nextcloud-rule-exclusions-plugin/9508171.yaml
@@ -6,7 +6,7 @@ meta:
   name: 9508171.yaml
 tests:
   - test_title: 9508171-1
-    desc: Entering password on a public share
+    desc: Entering password on a public share (Nextcloud changed the case on this endpoint, tested on 29.0.3)
     stages:
       - stage:
           input:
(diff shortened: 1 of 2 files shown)


tests/regression/nextcloud-rule-exclusions-plugin/9508171.yaml
```diff
@@ -0,0 +1,45 @@
...
+            port: 80
+            method: POST
+            uri: /s/random/authenticate/showshare
+            data: |
+              requesttoken=random&password=%3Cscript%3E&sharingToken=random&sharingType=3
+            version: HTTP/1.1
+          output:
+            no_log_contains: id "941101"
+  - test_title: 9508171-2
+    desc: Entering password on a public share
```
theseion, line 28:
```suggestion
    desc: Entering password on a public share (Nextcloud < v....)
```

theseion (changes requested):

azurit (approved):

azurit:
@theseion Is this ready to merge?

theseion (approved):

PR #86, opened 2024-07-19
refactor: overhaul and cleanup rule exclusions for settings

EsadCetiner:
Over the past year the Nextcloud plugin has grown dramatically in size with improved coverage and support for various features within Nextcloud, this has come at the expense of code readability and maintainability. Rule IDs have grown from 53 to 185 and the lines of code from 780 to 2756.

This PR is focused on overhauling personal/administration settings and has:

- Reduced the number of rules down to 181 (from 185)
- Rule exclusions have been reorganized to make it easier to find/modify/add rule exclusions relating to settings
- Comment clarity has been improved
- 6 Additional tests have been added

The Nextcloud plugin will likely remain large since Nextcloud incredibly feature rich, but it's a start.

PR #93, opened 2024-08-03
Revert "refactor: overhaul and cleanup rule exclusions for settings"

EsadCetiner:
Reverts coreruleset/nextcloud-rule-exclusions-plugin#86


theseion:
The failing tests are not due to your changes, but because `docker-compose` is an unknown command. Let me fix that.

EsadCetiner:
Ok good, I was having a heart attack. I ran the tests locally and they work just fine.

PR #94, opened 2024-08-16
refactor: cleanup and overhaul text editor rules and tests

EsadCetiner:
The rules for handling false positives for the markdown text editor were a bit messy and all over the place, especially the tests. The rules have now been cleaned up, simplified, and reorganized. This also fixes a false positive with the text editor that occurs at PL-2, it was easier to simplify the rules then add another rule.

- Real payloads are now being used for tests, this makes the tests smaller and should hopefully prevent any issues stemming from not using real payloads.
- About 100 lines of code has been removed.
- Rule count has been reduced from 184 down to 179.
- All rules for the markdown text editor are in one area instead of being split between file manager and text editor.


theseion:
Nice! I'll take a look soon.

plugins/nextcloud-rule-exclusions-before.conf
```diff
@@ -577,177 +537,82 @@ SecRule REQUEST_FILENAME "@contains /index.php/apps/files_texteditor/" \
     ctl:ruleRemoveTargetById=932150;ARGS:filename,\
     ctl:ruleRemoveTargetByTag=OWASP_CRS;ARGS:filecontents"
 
-# Keeping track of client session within a text editor
-# Nextcloud 28 and newer will include the document ID in the Request URI after session
+# Handle false positives for token related false positives (Used for session
+# tracking, multi-user editing, guest edits,)
+# There's either way too many parameters that are used for specific endpoints or the
```
theseion, line 542:
```suggestion
# There are either way too many parameters that are used for specific endpoints or the
```

theseion (changes requested):

theseion (approved):

PR #102, opened 2024-11-18
fix: renaming tagged faces in photos app

### LINKED ISSUE #100: Exclusion rules for Nextcloud-Recognize module incomplete
Exclusions rules for Nextcloud-Recognize Rule is not working correctly.
The recognise Modul allows renaming Photos to names instead of numbers.

The current rule works correctly if they are only addressed by numbers, but as soon as it is renamed to Names, the exclusion rule does not apply any more and the request is being blocked.

EsadCetiner:
closes #100


tests/regression/nextcloud-rule-exclusions-plugin/9508953.yaml
```diff
@@ -0,0 +1,51 @@
...
+                    <nc:hidden />
+                    <oc:favorite />
+                    <oc:fileid />
+                    <oc:permissions />
+                    <nc:nbItems />
+                  </d:prop>
+                </d:propfind>
+            version: HTTP/1.1
+          output:
+            no_log_contains: id "920420"
```
theseion, line 51: The rule removes `921110`.

theseion (changes requested):

PR #107, opened 2025-01-17
refactor: overhaul nc files, calendar, and contacts rule exclusions

EsadCetiner:
Rule exclusions for the File Manager is currently a mess and a pain to work with, these issues became even more obvious when I started work on fixing issues with the Nextcloud iOS app (The app is currently unusable) and I didn't want to compound the problem. I had to touch some rules for CalDav and CardDav since they both use similar endpoints (`remote.php`).

Since these changes are fairly big, I tested these changes against older Nextcloud versions and tested this in prod for a while so there shouldn't be any regressions. I added some tests where I could but this isn't an easy part of Nextcloud to add tests for.

Improvements:
- The lines of code has been reduced by about 250
- The number of rules has gone down from 184 to 168
- Rule exclusions for Nextcloud Files are no longer fragmented between `File Manager` and `Nextcloud Files`
- Due to the number of rule exclusions for the File Manager, I've split the rules into 3 subcategories, Settings, Web GUI, and file uploads/downloads.
- As a consequence of the refactor, a new false positive introduced in the Nextcloud 31 beta has also been fixed.
- Existing rules should be easier to maintain (Evident by the fix in Nextcloud 31 beta)


README.md
```diff
@@ -77,20 +77,6 @@ Apache libmodsecurity3 Example:
 </LocationMatch>
 ```
 
-## Relaxing file upload restrictions
```
theseion, line 80: Why did you remove this part from the readme?

plugins/nextcloud-rule-exclusions-before.conf
```diff
@@ -50,161 +50,180 @@ SecRule &TX:allowed_request_content_type "@eq 0" \
...
+# matches:
+#   /remote.php/dav/uploads/<username>/<random>/<random>
+#   /remote.php/dav/bulk
+#   /remote.php/dav/files/username/path/to/example
+#   /remote.php/webdav/path/to/file
+#   /public.php/webdav/path/to/file
+#   /public.php/dav/files/<share-id>/path/to/file
+#
+# NOTE: /remote.php/dav/webdav is no longer used for file uploads via
+# the web GUI as of NC 28
```
theseion, line 135: Here you no longer support a URL for NC <= 27, but three rules done you still support `if` for NC 26.

plugins/nextcloud-rule-exclusions-before.conf
```diff
@@ -765,9 +683,9 @@ SecRule REQUEST_FILENAME "@rx /remote\.php/dav/calendars/(?:[^/]+/){2}[^/]+\.ics
         "t:none,\
         ctl:ruleRemoveById=950100"
 
-# Editing/deleting existing calendar appointments
+# Editing/deleting an existing calendar appointments
 # Calendar description can be anything
-SecRule REQUEST_FILENAME "@rx /apps/calendar/v[0-9]/appointment_configs/[0-9]+$" \
+SecRule REQUEST_FILENAME "@rx /apps/calendar/v[0-9.]/appointment_configs/[0-9]+$" \
```
theseion, line 688: Is `v[0-9.]` correct? `v.` is a valid segment? If I had to guess, I'd say you're missing a `+` there: [v[0-9.]+`.

plugins/nextcloud-rule-exclusions-before.conf
```diff
@@ -778,8 +696,9 @@ SecRule REQUEST_FILENAME "@rx /apps/calendar/v[0-9]/appointment_configs/[0-9]+$"
     ctl:ruleRemoveTargetByTag=OWASP_CRS;ARGS:json.description,\
     setvar:'tx.allowed_methods=%{tx.allowed_methods} PUT DELETE'"
 
+# Creating an calendar appointment
 # Calendar description can be anything
-SecRule REQUEST_FILENAME "@rx /apps/calendar/v[0-9]/appointment_configs$" \
+SecRule REQUEST_FILENAME "@rx /apps/calendar/v[0-9.]/appointment_configs$" \
```
theseion, line 701: Same here: missing `+`?

plugins/nextcloud-rule-exclusions-before.conf
```diff
@@ -1041,7 +871,7 @@ SecRule REQUEST_FILENAME "@endsWith /settings/users" \
     ctl:ruleRemoveTargetByTag=OWASP_CRS;ARGS:password"
 
 # Nextcloud administrator setting/changing a user's password
-SecRule REQUEST_FILENAME "@rx /ocs/v[0-9]\.php/cloud/users(?:/[^/]+)?$" \
+SecRule REQUEST_FILENAME "@rx /ocs/v[0-9.]\.php/cloud/users(?:/[^/]+)?$" \
```
theseion, line 874: Same here: missing `+`?

plugins/nextcloud-rule-exclusions-before.conf
```diff
@@ -1359,7 +1189,7 @@ SecRule REQUEST_FILENAME "@rx /ocs/v[0-9.]+\.php/apps/provisioning_api/api/v[0-9
 
 # Personal Settings --> Appearance and Accessibility --> Dyslexia font
 # Disabling Dyslexia font
-SecRule REQUEST_FILENAME "@rx /ocs/v[0-9]\.php/apps/theming/api/v[0-9]/theme/opendyslexic$" \
+SecRule REQUEST_FILENAME "@rx /ocs/v[0-9.]\.php/apps/theming/api/v[0-9]/theme/opendyslexic$" \
```
theseion, line 1192: Same here: missing `+`?

plugins/nextcloud-rule-exclusions-before.conf
```diff
@@ -1532,7 +1362,7 @@ SecRule REQUEST_FILENAME "@rx /ocs/v[0-9]+\.php/profile/[^/]+$" \
 
 # Personal Settings --> Personal Info --> Website
 # When adding a link to your own user profile
-SecRule REQUEST_FILENAME "@rx /ocs/v[0-9]\.php/cloud/users/$" \
+SecRule REQUEST_FILENAME "@rx /ocs/v[0-9.]\.php/cloud/users/$" \
```
theseion, line 1365: Same here: missing `+`?

plugins/nextcloud-rule-exclusions-before.conf
```diff
@@ -1676,7 +1506,7 @@ SecRule REQUEST_FILENAME "@contains /apps/recognize/admin/settings/" \
 
 # Administration Settings --> Flow --> File retention & automatic deletion
 # Disabling automatic deletion of a tag type
-SecRule REQUEST_FILENAME "@rx /ocs/v[0-9]\.php/apps/files_retention/api/v[0-9]/retentions/[0-9]+$" \
+SecRule REQUEST_FILENAME "@rx /ocs/v[0-9.]\.php/apps/files_retention/api/v[0-9]/retentions/[0-9]+$" \
```
theseion, line 1509: Same here: missing `+`?

plugins/nextcloud-rule-exclusions-before.conf
```diff
@@ -1778,7 +1609,7 @@ SecRule REQUEST_FILENAME "@rx /ocs/v[0-9.]+\.php/apps/notifications/api/v[0-9.]+
     setvar:'tx.allowed_methods=%{tx.allowed_methods} DELETE'"
 
 # Sometimes a 502 error is returned when viewing/dismissing a notification
-SecRule REQUEST_FILENAME "@rx /ocs/v[0-9]\.php/apps/notifications/api/v[0-9]/notifications$" \
+SecRule REQUEST_FILENAME "@rx /ocs/v[0-9.]\.php/apps/notifications/api/v[0-9]/notifications$" \
```
theseion, line 1612: Same here: missing `+`?

plugins/nextcloud-rule-exclusions-before.conf
```diff
@@ -2020,7 +1851,7 @@ SecRule REQUEST_FILENAME "@endsWith /apps/dashboard/statuses" \
 #  /ocs/v2.php/cloud/<user group>/username/groups
 #  /ocs/v2.php/cloud/<user group>/username/disable
 #  /ocs/v2.php/cloud/<user group>/username/enable
-SecRule REQUEST_FILENAME "@rx /ocs/v[0-9]\.php/cloud/[^/]+/[^/]+(?:/groups|/disable|/enable)?$" \
+SecRule REQUEST_FILENAME "@rx /ocs/v[0-9.]\.php/cloud/[^/]+/[^/]+(?:/groups|/disable|/enable)?$" \
```
theseion, line 1854: Same here: missing `+`?

plugins/nextcloud-rule-exclusions-before.conf
```diff
@@ -2161,7 +1992,7 @@ SecRule REQUEST_FILENAME "@endsWith /apps/notes/settings" \
 #
 
 # Allow admin to add custom CSS code
-SecRule REQUEST_FILENAME "@rx /ocs/v[0-9]\.php/apps/provisioning_api/api/v[0-9]/config/apps/theming_customcss/customcss$" \
+SecRule REQUEST_FILENAME "@rx /ocs/v[0-9.]\.php/apps/provisioning_api/api/v[0-9]/config/apps/theming_customcss/customcss$" \
```
theseion, line 1995: Same here: missing `+`?

plugins/nextcloud-rule-exclusions-before.conf
```diff
@@ -2177,7 +2008,7 @@ SecRule REQUEST_FILENAME "@rx /ocs/v[0-9]\.php/apps/provisioning_api/api/v[0-9]/
 # When using Nextcloud circles
 # PUT - When editing/creating a circle
 # DELETE - When deleting an existing circle
-SecRule REQUEST_FILENAME "@rx /ocs/v[0-9]\.php/apps/circles/circles/" \
+SecRule REQUEST_FILENAME "@rx /ocs/v[0-9.]\.php/apps/circles/circles/" \
```
theseion, line 2011: Same here: missing `+`?

plugins/nextcloud-rule-exclusions-before.conf
```diff
@@ -2187,7 +2018,7 @@ SecRule REQUEST_FILENAME "@rx /ocs/v[0-9]\.php/apps/circles/circles/" \
     setvar:'tx.allowed_methods=%{tx.allowed_methods} PUT DELETE'"
 
 # When setting a unique password for all shares to a circle
-SecRule REQUEST_FILENAME "@rx /ocs/v[0-9]\.php/apps/circles/circles/[^/]+/setting$" \
+SecRule REQUEST_FILENAME "@rx /ocs/v[0-9.]\.php/apps/circles/circles/[^/]+/setting$" \
```
theseion, line 2021: Same here: missing `+`?

plugins/nextcloud-rule-exclusions-before.conf
```diff
@@ -2202,7 +2033,7 @@ SecRule REQUEST_FILENAME "@rx /ocs/v[0-9]\.php/apps/circles/circles/[^/]+/settin
 #
 
 # When changing settings for Nextcloud photos
-SecRule REQUEST_FILENAME "@rx /apps/photos/api/v[0-9]/config/config/(?:croppedLayout|photosLocation)$" \
+SecRule REQUEST_FILENAME "@rx /apps/photos/api/v[0-9.]/config/config/(?:croppedLayout|photosLocation)$" \
```
theseion, line 2036: Same here: missing `+`?

theseion (changes requested):
Good stuff!

plugins/nextcloud-rule-exclusions-before.conf
EsadCetiner: Not sure if I understand you fully, but that endpoint is included in the regex below. It's a bit messy to read but I don't see a better way of handling this false positive because of the content type issue.

theseion: Your regex does not match `/remote.php/dav/webdav`. From the rule comment I had expected that, but now it sounds like you actually meant to match it too...? Either way, I believe regex should match `/remote.php/dav/webdav` and the URL should be part of the list above. You can still keep the deprecation comment, that's good information to have.

EsadCetiner: Yes it was meant to match `/remote.php/dav/webdav`, I noticed a few more issues so I fixed those as well.

theseion (approved):

PR #109, opened 2025-01-21
refactor: globally allow all webdav methods

EsadCetiner:
Nextcloud makes very heavy use of WebDAV methods, it's pretty much used everywhere even in the Web GUI. So much so about 1/3 of all rules were solely for allowing WebDAV methods. Now with this refactor, these methods are allowed globally within the plugin. This should be an big improvement in terms of maintenance and usability for the plugin.

I didn't test this for very long as existing test coverage should pick up any regressions, and this refactor is relatively simple. 

- Reduced the number of lines of code by about 700.
- Number of rules has gone down from 169 to 103.
- A nice bonus is that some unsupported Nextcloud Apps may now be usable for more basic use cases at PL-1.
- The plugin will require significantly less maintenance in the long run, adding support for additional apps in the future will be easier too.
- Fixes a few false positives in supported apps thanks to the refactor.


plugins/nextcloud-rule-exclusions-before.conf
```diff
@@ -414,24 +348,19 @@ SecRule REQUEST_FILENAME "@rx /ocs/v[0-9.]+\.php/search/providers(?:/[^/]+/searc
...
-# Calendars: REPORT
-# Others in the code or js files: PATCH MKCOL MOVE TRACE
-# Others that I added just in case, and they seem related:
-#   CHECKOUT COPY LOCK MERGE MKACTIVITY UNLOCK.
-SecRule REQUEST_FILENAME "@rx /(?:remote|index|public)\.php/" \
+# Nextcloud makes very heavy use of WebDAV request methods, this
+# rule allows all of the required request methods for Nextcloud globally.
+# Only allowing the exact request methods for the exact endpoints will result
+# in a huge explosion of the number of rule exclusions and time spent maintaining these rules.
+# NOTE: sometimes dav clients will make requests to `/remote.php/dav` with no trailing slash
```
theseion, line 355:
```suggestion
# NOTE: sometimes DAV clients will make requests to `/remote.php/dav` with no trailing slash
```

plugins/nextcloud-rule-exclusions-before.conf
theseion, line 1: I don't see where you allow those methods by default...

theseion (changes requested):

plugins/nextcloud-rule-exclusions-before.conf
EsadCetiner: 9508190 is doing approximately that, I didn't use the `@unconditionalMatch` operator just in case somebody has this plugin activated for something other than Nextcloud. I can change the operator if you want and delete the tests for that rule.

theseion: I see. So you basically just expanded the regex to include more URLs, right?

EsadCetiner: Yes, it should cover pretty much all of Nextcloud's URLs proximately.

theseion (approved):

PR #111, opened 2025-02-06
fix: add workaround for ios file upload issue

EsadCetiner:
This PR adds a workaround for the Nextcloud iOS app not setting the correct content type when uploading files making it impossible for CRS to parse it (Pretty much every single CRS rule is triggered). Since I've essentially disabled all rules for the request body (Excluding the `REQUEST_BODY` variable) I fear that this might cause something similar to [CVE-2021-35368](https://nvd.nist.gov/vuln/detail/CVE-2021-35368) so I've disabled the workaround by default but given an option to enable the workaround for those that need it. I've done what I could to minimize the risk/impact of a bypass for those that need this functionality, but I'm not fully satisfied with the solution.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ Apache with ModSecurity2:
 </LocationMatch>
 ```
 
-NGINX with libModSecurity3:
+nginx with libModSecurity3:
 
 ```
 location ~ (?:/index\.php/apps/files/ajax/upload\.php|/remote\.php/dav/(?:bulk|files/|uploads/)) { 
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -57,55 +57,44 @@ Below you can find a list of known limitations along with workarounds for this i
...
 ```
 <LocationMatch "(?:/index\.php/apps/files/ajax/upload\.php|/remote\.php/dav/(?:bulk|files/|uploads/))">
-    modsecurity_rules 'SecRequestBodyLimit 1073741824'
+    SecRequestBodyLimit 10737418240
+    SecRequestBodyNoFilesLimit 10737418240
 </LocationMatch>
 ```
 
-### Increasing max request body size
+NGINX with libModSecurity3:
```
theseion, line 75:
```suggestion
nginx with libModSecurity3:
```

README.md
```diff
@@ -57,55 +57,44 @@ Below you can find a list of known limitations along with workarounds for this i
...
 
-Apache libmodsecurity3 Example:
-```
-<location "/remote.php/dav/files/">
-   modsecurity_rules 'SecRequestBodyNoFilesLimit 144384'
-</location>
-```
+### Fixing file uploads on iOS
+
+The Nextcloud iOS app does not set the correct content type when uploading files, which causes issues with CRS as it relies on the content type to correctly parse data.
```
theseion, line 94:
```suggestion
The Nextcloud iOS app does not set the correct content type when uploading files, which causes issues with ModSecurity as it relies on the content type to correctly parse data.
```

theseion (changes requested):
Makes sense to me, especially if there's a good chance that the issue gets fixed.

theseion (approved):

PR #114, opened 2025-04-09
fix: add workaround for libmodsecurity3 bug

EsadCetiner:
Adds a workaround for https://github.com/owasp-modsecurity/ModSecurity/issues/3356 when setting `SecRequestBodyNoFilesLimit` to a high value for libModSecurity3 users. This workaround has only been tested against the configuration recommendations in the plugin readme.

### DIFF
--- a/plugins/nextcloud-rule-exclusions-before.conf
+++ b/plugins/nextcloud-rule-exclusions-before.conf
@@ -897,6 +897,29 @@ SecRule REQUEST_FILENAME "@rx (?i)\.(?:avif|bmp|brotli|bvr|css|eot|gif|gz|ico|jp
     ver:'nextcloud-rule-exclusions-plugin/1.3.1',\
     ctl:ruleRemoveTargetById=932236;ARGS:v"
 
+# Setting SecRequestBodyNoFilesLimit to a very high value results in all requests
+# being interpreted as exceeding the configured limit due to a bug in libModSecurity3.
+# This rule is a workaround, it's possible this bug happens in other areas
+# but this has only been tested based on the recommendations in the plugin's readme.
+# See: https://github.com/owasp-modsecurity/ModSecurity/issues/3356
+SecRule REQUEST_FILENAME "@contains /remote.php/dav/files" \
+    "id:9508512,\
+    phase:1,\
+    pass,\
+    t:none,\
+    nolog,\
+    ver:'nextcloud-rule-exclusions-plugin/1.3.1',\
+    chain"
+    SecRule REQUEST_METHOD "@pm MKCOL DELETE" \
+        "t:none,\
+        chain"
+        SecRule REQUEST_HEADERS:Content-Type "@streq application/x-www-form-urlencoded" \
+            "t:none,\
+            chain"
+            SecRule REQUEST_HEADERS:content-length "@eq 0" \
+                "t:none,\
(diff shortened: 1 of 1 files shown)


plugins/nextcloud-rule-exclusions-before.conf
```diff
@@ -897,6 +897,29 @@ SecRule REQUEST_FILENAME "@rx (?i)\.(?:avif|bmp|brotli|bvr|css|eot|gif|gz|ico|jp
     ver:'nextcloud-rule-exclusions-plugin/1.3.1',\
     ctl:ruleRemoveTargetById=932236;ARGS:v"
 
+# Setting SecRequestBodyNoFilesLimit to a very high value results in all requests
+# being interpreted as exceeding the configured limit due to a bug in libModSecurity3.
+# This rule is a workaround, it's possible this bug happens in other areas
```
theseion, line 902:
```suggestion
# This rule is a workaround for this bug. It's possible this bug happens in other areas
```

plugins/nextcloud-rule-exclusions-before.conf
```diff
@@ -897,6 +897,29 @@ SecRule REQUEST_FILENAME "@rx (?i)\.(?:avif|bmp|brotli|bvr|css|eot|gif|gz|ico|jp
...
+    nolog,\
+    ver:'nextcloud-rule-exclusions-plugin/1.3.1',\
+    chain"
+    SecRule REQUEST_METHOD "@pm MKCOL DELETE" \
+        "t:none,\
+        chain"
+        SecRule REQUEST_HEADERS:Content-Type "@streq application/x-www-form-urlencoded" \
+            "t:none,\
+            chain"
+            SecRule REQUEST_HEADERS:content-length "@eq 0" \
```
theseion, line 919: I don't understand the logic of the chain. Why does the `Content-Length` header need to have the value `0`?. Please explain in the comment what this rule is looking for and why.

theseion (changes requested):

plugins/nextcloud-rule-exclusions-before.conf
EsadCetiner: The comment I wrote is meant to be generic. The only false positive I encountered were with the MKCOL and DELETE methods, normally those methods don't have any request body but a content type/length were set anyway. Pretty common behavior with Nextcloud for some reason.

theseion: I see.

plugins/nextcloud-rule-exclusions-before.conf
```diff
@@ -897,6 +897,29 @@ SecRule REQUEST_FILENAME "@rx (?i)\.(?:avif|bmp|brotli|bvr|css|eot|gif|gz|ico|jp
     ver:'nextcloud-rule-exclusions-plugin/1.3.1',\
     ctl:ruleRemoveTargetById=932236;ARGS:v"
 
+# Setting SecRequestBodyNoFilesLimit to a very high value results in all requests
+# being interpreted as exceeding the configured limit due to a bug in libModSecurity3.
+# This rule is a workaround for this bug. It's possible this bug happens in other areas
```
theseion, line 902:
```suggestion
# This rule works around the bug for some common requests. It's possible this bug happens in other areas
```

Issue #122, opened 2025-05-14
False positive when updating calendar with DAVx5


ti-guru:
Hi,

I've set up the nextcloud rule exclusions with caddy and coraza running in nested Docker containers in LXD. When creating a calendar event on my android phone and syncing it to the server through [DAVx5](https://play.google.com/store/apps/details?id=at.bitfire.davdroid&pli=1) I get the following error logged to console:
```
caddy-1  | {"level":"error","ts":1747254609.1309803,"logger":"http.handlers.waf","msg":"[client \"127.0.0.1\"] Coraza: Warning. HTTP Header Injection Attack via payload (CR/LF detected) [file \"@owasp_crs/REQUEST-921-PROTOCOL-ATTACK.conf\"] [line \"4777\"] [id \"921150\"] [rev \"\"] [msg \"HTTP Header Injection Attack via payload (CR/LF detected)\"] [data \"Matched Data: \\r found within ARGS_NAMES:BEGIN:VCALENDAR\\r\\nVERSION:2.0\\r\\nPRODID:DAVx5/4.4.10-gplay ical4j/3.2.19\\r\\nBEGIN:VEVENT\\r\\nDTSTAMP:20250514T203008Z\\r\\nUID:<longUID string>\\r\\nSUMMARY:Test\\r\\nDTSTART;TZID: BEGIN:VCALENDAR\\r\\nVERSION:2.0\\r\\nPRODID:DAVx5/4.4.10-gplay ica\"] [severity \"critical\"] [ver \"OWASP_CRS/4.7.0\"] [maturity \"0\"] [accuracy \"0\"] [tag \"application-multi\"] [tag \"language-multi\"] [tag \"platform-multi\"] [tag \"attack-protocol\"] [tag \"paranoia-level/1\"] [tag \"OWASP_CRS\"] [tag \"capec/1000/210/272/220/33\"] [hostname \"\"] [uri \"/remote.php/dav/calendars/<username>/<calendar name>/<calendar event id>.ics\"] [unique_id \"VnlVOaXIbllaCyMR\"]"}
[...]

PR #123, opened 2025-05-16
docs: document issue with coraza default body parser

EsadCetiner:
By default Coraza activates the `URLENCODED` body parser when no other parser has been configured, this behavior can cause issues when uploading files in Nextcloud.

resolves: #122

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,7 +139,7 @@ SecRule REMOTE_ADDR "@ipMatch your-server-ip" \
 
 ### Coraza default body processor
 
-By default Coraza will use the `URLENCODED` body processor when no other request body parser has been configured, what body parser Coraza uses is defined by the `coraza.conf` recommended file. This behavior is not desirable for Nextcloud since Nextcloud is used to upload files which may include raw binary data or pretty much anything. This behavior can be changed by using the `RAW` body processor when no other parser has been configured.
+By default, Coraza will use the `URLENCODED` body processor when no other request body processor has been configured. Which body processor Coraza uses is defined in the `coraza.conf` file. The default behavior is not desirable for Nextcloud since Nextcloud is used to upload files which may include raw binary data or pretty much anything. The following rule can be used to enable the `RAW` body processor when no other parser has been configured.
 
 ```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -137,6 +137,21 @@ SecRule REMOTE_ADDR "@ipMatch your-server-ip" \
     ctl:ruleRemoveTargetById=922130;MULTIPART_PART_HEADERS"
 ```
 
+### Coraza default body processor
+
+By default Coraza will use the `URLENCODED` body processor when no other request body parser has been configured, what body parser Coraza uses is defined by the `coraza.conf` recommended file. This behavior is not desirable for Nextcloud since Nextcloud is used to upload files which may include raw binary data or pretty much anything. This behavior can be changed by using the `RAW` body processor when no other parser has been configured.
```
theseion, line 142:
```suggestion
By default, Coraza will use the `URLENCODED` body processor when no other request body processor has been configured. Which body processor Coraza uses is defined in the `coraza.conf` file. The default behavior is not desirable for Nextcloud since Nextcloud is used to upload files which may include raw binary data or pretty much anything. The following rule can be used to enable the `RAW` body processor when no other parser has been configured.
```

README.md
```diff
@@ -137,6 +137,21 @@ SecRule REMOTE_ADDR "@ipMatch your-server-ip" \
...
+
+```
+SecRule REQBODY_PROCESSOR "!@rx (?:URLENCODED|MULTIPART|XML|JSON)" \
+    "id:9508033,\
```
theseion, line 146: Is this the ID you want users to use? I'm pretty sure that users will simply copy the rule and the ID will stick. So it should be an ID that you won't use anywhere else accidentally.

theseion (changes requested):

README.md
EsadCetiner: Yes it should be fine, the first 100 rules of plugins are reserved for initialization and configuration. These rule ids are pretty much never touched within the plugin so I think it's fine to use them within the readme.

theseion (approved):

PR #139, opened 2025-12-06
fix: response rule exclusion disabling incorrect target

EsadCetiner:
Excluded target for response rule was `REQUEST_BODY` instead of `RESPONSE_BODY`


fzipi (approved):
Makes sense.

PR #152, opened 2026-05-03
fix: iOS cliient sending post request without content length/transfer-encoding during configuration

theSCHAPPY:
* During ios client configuration a POST request with insufficient content length will trigger mod_security

PR #153, opened 2026-05-08
Fix for iOS app setup

theSCHAPPY:
New PR using signed commits of rule `9508503` for fixing iOS app polling triggering rule `920180` (in response to #152, which was unable to merge due to unsigned commits).


theSCHAPPY:
@EsadCetiner Please revoke this PR, it is replaced by #154. Thx!

PR #154, opened 2026-05-08
Fix ios setup

theSCHAPPY:
Added rule 9508503 to handle iOS client polling during setup triggering 920180. Follow up to #152 and #153 using unsigned commits blocking merge.


theSCHAPPY:
@EsadCetiner Finally, all commits are recognized as signed commits.

EsadCetiner (approved):

EsadCetiner:
@theSCHAPPY Thanks, I can merge now.

PR #162, opened 2026-07-31
fix(collabora-admin): loading font file in admin panel

EsadCetiner:


plugins/nextcloud-rule-exclusions-before.conf
```diff
@@ -1324,6 +1324,20 @@ SecRule REQUEST_FILENAME "@contains /cool/https://" \
         t:urlDecode,\
         ctl:ruleRemoveTargetById=931131;REQUEST_FILENAME"
 
+# Font file is sent compressed without a Content-Encoding header, therefore respone
```
theseion:
```suggestion
# Font file is sent compressed without a Content-Encoding header, therefore response
```

tests/regression/nextcloud-rule-exclusions-plugin/9508559.yaml
```diff
@@ -0,0 +1,46 @@
+---
+meta:
+  author: "Esad Cetiner"
+  description: "Nextcloud Office: Loading font file for Collabora Admin panel"
+rule_id: 9508559
+tests:
+
+      - input:
```
theseion, line 9:
```suggestion
tests:
  - test_id: 1
    stages:
      - input:
```

theseion (changes requested):

theseion (approved):

PR #168, opened 2026-09-13
feat: add plugin.yaml descriptor

fzipi:
## Summary
- Adds `plugin.yaml`, the machine-readable plugin descriptor, per coreruleset/plugin-registry#21 and validated against coreruleset/plugin-registry's `plugin-schema.json` (schema_version 1).

## Test plan
- [x] Validated with `check-jsonschema --schemafile plugin-schema.json plugin.yaml`

🤖 Generated with [Claude Code](https://claude.com/claude-code)


plugin.yaml
```diff
@@ -0,0 +1,54 @@
...
+      default: 1
+      description: "Enable or disable the Nextcloud rule exclusions plugin (0 to disable)"
+    - name: "tx.nextcloud-rule-exclusions-plugin_ios_workaround_enabled"
+      type: "boolean"
+      default: 0
+      description: "Enable the workaround for the Nextcloud iOS app's file upload content type issue (disabled by default due to bypass risk)"
+    - name: "tx.nextcloud-rule-exclusions-plugin_server_crawler_rules_enabled"
+      type: "boolean"
+      default: 0
+      description: "Enable rule exclusions for the Nextcloud Server Crawler (disabled by default; also requires setting the crawler's IP address)"
```
EsadCetiner: The implementation of this feature is a bit hacky due to engine limitations, the intended use case is to only enable the boolean for a specific IP addresses.
```suggestion
      description: "Enable rule exclusions for the Nextcloud Server Crawler (disabled by default; Should only be enabled via a SecRule scoped to server's IP address)"
```

EsadCetiner (changes requested):

EsadCetiner (approved):

